PR #29, opened 2024-01-09
(Feat): Machine Identity Auth support

DanielHougaard:
- Add machine identity support
- Fix error responses all having the same message when resource's fail
- Fix slack link in certain error messages

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -117,9 +117,6 @@ func (client Client) CallDeleteSecretsV3(request DeleteSecretV3Request) error {
 
 func (client Client) CallUpdateSecretsV3(request UpdateSecretByNameV3Request) error {
 
-	//var secretName = request.SecretName
-	//request.SecretName = ""
-
 	var secretsResponse GetEncryptedSecretsV3Response
 	response, err := client.Config.HttpClient.
 		R().
--- a/client/api.go
+++ b/client/api.go
@@ -6,6 +6,30 @@ import (
 
 const USER_AGENT = "terraform"
 
+func (client Client) UniversalMachineIdentityAuth() (string, error) {
+	if client.Config.ClientId == "" || client.Config.ClientSecret == "" {
+		return "", fmt.Errorf("you must set the client secret and client ID for the client before making calls")
+	}
+
+	var loginResponse UniversalMachineIdentityAuthResponse
+
+	res, err := client.Config.HttpClient.R().SetResult(&loginResponse).SetHeader("User-Agent", USER_AGENT).SetBody(map[string]string{
+		"clientId":     client.Config.ClientId,
+		"clientSecret": client.Config.ClientSecret,
+	}).Post("api/v1/auth/universal-auth/login")
+
+	if err != nil {
+		return "", fmt.Errorf("UniversalMachineIdentityAuth: Unable to complete api request [err=%s]", err)
+	}
+
+	if res.IsError() {
+		return "", fmt.Errorf("UniversalMachineIdentityAuth: Unsuccessful response: [response=%s]", res)
+	}
+
+	return loginResponse.AccessToken, nil
+}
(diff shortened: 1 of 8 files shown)


client/api.go
```diff
@@ -93,6 +116,10 @@ func (client Client) CallDeleteSecretsV3(request DeleteSecretV3Request) error {
 }
 
 func (client Client) CallUpdateSecretsV3(request UpdateSecretByNameV3Request) error {
+
+	//var secretName = request.SecretName
+	//request.SecretName = ""
+
```
maidul98, line 122: we shoud get rid of these if we don't need it

client/client.go
```diff
@@ -24,14 +43,32 @@ func NewClient(cnf Config) (*Client, error) {
 		cnf.HttpClient.SetBaseURL(cnf.HostURL)
 	}
 
-	if cnf.ServiceToken == "" {
-		return nil, fmt.Errorf("you must set the service token for the client before making calls")
+	if cnf.ServiceToken == "" && cnf.ClientId == "" && cnf.ClientSecret == "" {
+		return nil, fmt.Errorf("you must set the service token, or a client secret and client ID for the client before making calls")
```
maidul98, line 47: Let's mention the auth strategy name 

```
you must configure a authentication method such as service tokens or Universal Auth before making calls
```

client/client.go
```diff
@@ -24,14 +43,32 @@ func NewClient(cnf Config) (*Client, error) {
...
+		token, err := Client{cnf}.UniversalMachineIdentityAuth()
+
+		if err != nil {
+			return nil, fmt.Errorf("unable to authenticate with universal machine identity [err=%s]", err)
+		}
+
+		authToken = token
+		cnf.AuthStrategy = AuthStrategy.UNIVERSAL_MACHINE_IDENTITY
+	}
+	if cnf.ServiceToken != "" && authToken == "" {
```
maidul98, line 62: I think we should add a check to make sure they aren't trying to set multiple auth methods at once. What do you think?

client/secret-operations.go
```diff
@@ -52,3 +52,27 @@ func (client Client) GetPlainTextSecretsViaServiceToken(secretFolderPath string,
 
 	return plainTextSecrets, &serviceTokenDetails, nil
 }
+
+func (client Client) GetRawSecretsViaMachineIdentity(secretFolderPath string, envSlug string, workspaceId string) ([]RawV3Secret, error) {
```
maidul98, line 56: Nit: let's keep the name more generic since raw ends points aren't only bound to machine identity. I think GetRawSecrets would work

infisical/provider/provider.go
```diff
@@ -80,9 +94,15 @@ func (p *infisicalProvider) Configure(ctx context.Context, req provider.Configur
...
 	}
 
-	serviceToken := os.Getenv("INFISICAL_SERVICE_TOKEN")
 	host := os.Getenv("INFISICAL_HOST")
 
+	// Service Token
+	serviceToken := os.Getenv("INFISICAL_SERVICE_TOKEN")
+
+	// Machine Identity
+	clientId := os.Getenv("INFISICAL_CLIENT_ID")
```
maidul98, line 103: This should be specifc to the auth name. We should include UNIVERSAL_AUTH in the name. Something like `INFISICAL_UNIVERSAL_AUTH_CLIENT_ID` would be good. Same for client secret

maidul98 (commented):
Left some minor feedback. Should be good to go after. Nice work

client/client.go
DanielHougaard: I agree. I refactored the client auth code to be more readable and gave it a better validation flow

maidul98 (approved):

maidul98 (approved):

PR #30, opened 2024-01-14
[feat] add support to reference secrets in terraform

jithenms:
Add support to reference secrets in terraform through interpolation syntax that is supported in CLI and native integrations

*Committed the tests for reviewers to view but upon approval will clean up

Closes [#1277](https://github.com/Infisical/infisical/issues/1277)


maidul98:
reference support has already been added, please update your operator to the latest version

PR #34, opened 2024-02-27
Feat: Add projects to Terraform Provider

DanielHougaard:
This PR introduces projects support for the Infisical Terraform Provider.

On the data source you can access environments by the environment slug like seen below. 

You access environments like so, `infisical_projects.test-project.environments["dev"]`. The reason for making it accessible like an object rather than on an array-basis, is to improve the user experience when interacting with environments. This also conveniently fits our secrets implementation.

⚠️ Do not merge/release before the V2 project-slug based router (infisical/infisical#1449) is released.


client/api.go
```diff
@@ -268,3 +268,89 @@ func (client Client) CallGetSingleRawSecretByNameV3(request GetSingleSecretByNam
...
+		SetHeader("User-Agent", USER_AGENT).
+		SetBody(request).
+		Post("api/v2/workspace")
+
+	if err != nil {
+		return CreateProjectResponse{}, fmt.Errorf("CallCreateProject: Unable to complete api request [err=%s]", err)
+	}
+
+	if response.IsError() {
+		return CreateProjectResponse{}, fmt.Errorf("CallCreateProject: Unsuccessful response. [response=%s]", response)
```
maidul98, line 294: can we make error logs format this way please? It would provide more details for users when they get error https://github.com/Infisical/infisical/blob/main/cli/packages/api/api.go#L29

docs/resources/project.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `name` (String) The name of the project
+- `organization_id` (String) The ID of the organization to which the project belongs
+- `slug` (String) The slug of the project
```
maidul98, line 57: I thought slug is optional? It is under required

maidul98 (commented):
Looks solid, really good work! Once we get the API PR merged, i'll try this out manually

client/api.go
DanielHougaard: Slack thread 🧵

Marking as resolved as discussed.

maidul98 (approved):
lgtm!

PR #35, opened 2024-05-21
project user management

akhilmhdh:
1. Added support for project user management
2. Option to invite users on project creation
3. Invite, Update and Remove users and there roles from a project.
4. Add, Update and Remove identity and there roles from a project.


maidul98 (changes requested):
Sent all feedback via slack

maidul98 (approved):
lgtm

PR #36, opened 2024-06-02
feature: Using terraform together with infisical login.

akhilmhdh:
This PR enabled users to use terraform with Infisical CLI. It checks for machine identity, then service token and then finally checks for the keyring saved from infisical CLI. 

User's must pass a profile argument or env `INFISICAL_USER_PROFILE` to ensure the right one is picked from keyring. (In future when you have multiple profiles in via cli login).

This improves ux as user's don't have to maintain the identity token for each project user and also good audit log.


varonix0 (changes requested):
Left two comments! 

I think this looks good, two things 
1. As you mentioned yourself, we should rebase this with main before merging
2. I think it's a good idea to just double-check and test that everything is working with user-based and machine identity-based auth

maidul98:
too stale

PR #48, opened 2024-07-23
misc: added support for OIDC auth

sheensantoscapadngan:
this PR adds support for OIDC auth.

With this, platforms like Spacelift can now be used without hardcoded tokens (via OIDC Auth)
Similar usage:
https://docs.spacelift.io/integrations/cloud-providers/oidc/vault-oidc

Auth configuration after these changes:
image
image

This schema update was made in a non-breaking way


internal/client/client.go
```diff
@@ -66,6 +79,15 @@ func NewClient(cnf Config) (*Client, error) {
...
 		cnf.AuthStrategy = AuthStrategy.SERVICE_TOKEN
+	} else if usingNativeAuth {
+		token, err := Client{cnf}.OidcMachineIdentityAuth()
+
+		if err != nil {
+			return nil, fmt.Errorf("unable to authenticate with oidc machine identity [err=%s]", err)
+		}
+
+		cnf.HttpClient.SetAuthToken(token)
+		cnf.AuthStrategy = AuthStrategy.UNIVERSAL_MACHINE_IDENTITY
```
sheensantoscapadngan, line 90: the reuse here is intentional... in the end the schema is the same because OIDC auth is machine identity and machine identity is a single category

internal/client/client.go
```diff
@@ -15,22 +15,26 @@ type AuthStrategyType string
...
-	AuthStrategy AuthStrategyType
+	AuthStrategy          AuthStrategyType
+	IsMachineIdentityAuth bool
 
 	// Service Token Auth
 	ServiceToken string
 
 	// Universal Machine Identity Auth
 	ClientId     string
 	ClientSecret string
```
DanielHougaard, line 36: Nit: Ideally I think we should keep the credentials as an interface on the config, so we avoid having a cluttered config. We do something along the lines of this in the Infisical Agent.

I can take a deeper look into this once we add more auth methods though!

internal/client/login.go
```diff
@@ -43,3 +46,32 @@ func (client Client) GetServiceTokenDetailsV2() (GetServiceTokenDetailsResponse,
...
+	if authJwt == "" {
+		return "", fmt.Errorf("INFISICAL_AUTH_JWT is not present in the environment")
+	}
+
+	var loginResponse OidcMachineIdentityAuthResponse
+
+	res, err := client.Config.HttpClient.R().SetResult(&loginResponse).SetHeader("User-Agent", USER_AGENT).SetBody(map[string]string{
+		"identityId": client.Config.IdentityId,
+		"jwt":        authJwt,
+	}).Post("api/v1/auth/oidc-auth/login")
```
DanielHougaard, line 66: Any reason for not using our Go SDK?

internal/client/model.go
```diff
@@ -193,6 +193,13 @@ type UniversalMachineIdentityAuthResponse struct {
 	TokenType         string `json:"tokenType"`
 }
 
+type OidcMachineIdentityAuthResponse struct {
```
DanielHougaard, line 196: I think we can just combine `UniversalMachineIdentityAuthResponse` and `OidcMachineIdentityAuthResponse` into a single struct called `MachineIdentityAuthResponse`

internal/provider/provider.go
```diff
@@ -77,6 +92,39 @@ func (p *infisicalProvider) Schema(ctx context.Context, _ provider.SchemaRequest
...
 			},
+			"auth": schema.SingleNestedAttribute{
+				Optional:    true,
+				Description: "The configuration values for authentication",
+				Attributes: map[string]schema.Attribute{
+					"universal": schema.SingleNestedAttribute{
+						Optional:    true,
+						Description: "The configuration values for Universal Auth",
+						Attributes: map[string]schema.Attribute{
+							"client_id": schema.StringAttribute{
```
DanielHougaard, line 103: Should `client_id` and `client_secret` be `optional: true`?

internal/provider/provider.go
```diff
@@ -77,6 +92,39 @@ func (p *infisicalProvider) Schema(ctx context.Context, _ provider.SchemaRequest
...
+								Description: "Machine identity client secret. Used to fetch/modify secrets for a given project",
+							},
+						},
+					},
+					"oidc": schema.SingleNestedAttribute{
+						Optional:    true,
+						Description: "The configuration values for OIDC Auth",
+						Attributes: map[string]schema.Attribute{
+							"identity_id": schema.StringAttribute{
+								Optional:    true,
```
DanielHougaard, line 120: Should `identity_id` be `optional: true`?

internal/provider/provider.go
```diff
@@ -104,6 +152,7 @@ func (p *infisicalProvider) Configure(ctx context.Context, req provider.Configur
 	// Machine Identity
 	clientId := os.Getenv("INFISICAL_UNIVERSAL_AUTH_CLIENT_ID")
 	clientSecret := os.Getenv("INFISICAL_UNIVERSAL_AUTH_CLIENT_SECRET")
+	identityId := os.Getenv("INFISICAL_MACHINE_IDENTITY_ID")
```
DanielHougaard, line 155: I think it makes sense to create a file with constants (`constants.go`), if we don't have one already. We can put all these static strings like `INFISICAL_MACHINE_IDENTITY_ID` in there, so we don't mix them up if we need to use them elsewhere

DanielHougaard (changes requested):
Good work! I left a few comments for you :-)

internal/client/model.go
sheensantoscapadngan: yup I agree

internal/provider/provider.go
sheensantoscapadngan: ah right good catch, they should be Required

internal/provider/provider.go
sheensantoscapadngan: same here

internal/provider/provider.go
sheensantoscapadngan: hmmm for this one, we're currently allowing users to define client_id and client_secret via ENV rather than directly in the configuration which is why this has to be optional

DanielHougaard (approved):
lgtm!

PR #49, opened 2024-08-06
Infisical import secret resource added.

swapnilkuwar:
This PR adds the Secret Import Resource.

I have also added validation that if the secret import is deleted outside the terraform then the `terraform plan` should update the existing state instead of throwing an error.

I am using the secret import list API to fetch the single secret import as we don't have the get secret import by id API.


DanielHougaard (approved):
great work!

PR #50, opened 2024-09-10
add secret reminder attribute

meetcshah19:
This PR adds secret reminder attribute to the secret resource.

Example usage:
```
resource "infisical_secret" "smtp_secret" {
  name         = "SMTP"
  value        = "<some key>"
  env_slug     = "dev"
  workspace_id = "PROJECT_ID"
  folder_path  = "/mail-service"
  secret_reminder = {
    note        = "Rotate this secret using X API"
    repeat_days = 30
  }
}
```


internal/client/model.go
```diff
@@ -367,21 +367,23 @@ type EncryptedSecret struct {
 
 // create secrets.
 type CreateSecretV3Request struct {
```
maidul98, line 369: Since we are not supporting reminders for end to end encrypted (e2ee) endpoints, you can remove these new fields you added

internal/provider/resource/secret_resource.go
```diff
@@ -29,15 +29,21 @@ type secretResource struct {
...
-	EnvSlug     types.String `tfsdk:"env_slug"`
-	Name        types.String `tfsdk:"name"`
-	Value       types.String `tfsdk:"value"`
-	WorkspaceId types.String `tfsdk:"workspace_id"`
-	LastUpdated types.String `tfsdk:"last_updated"`
-	Tags        types.List   `tfsdk:"tag_ids"`
+	FolderPath     types.String    `tfsdk:"folder_path"`
+	EnvSlug        types.String    `tfsdk:"env_slug"`
+	Name           types.String    `tfsdk:"name"`
+	SecretReminder *SecretReminder `tfsdk:"secret_reminder"`
```
maidul98, line 42: You don't need the `*SecretReminder`. You can just refer to the type like so `SecretReminder SecretReminder `tfsdk:"secret_reminder"`

internal/provider/resource/secret_resource.go
```diff
@@ -85,6 +91,22 @@ func (r *secretResource) Schema(_ context.Context, _ resource.SchemaRequest, res
 				Optional:    true,
 				Description: "Tag ids to be attached for the secrets.",
 			},
+			"secret_reminder": schema.SingleNestedAttribute{
+				Attributes: map[string]schema.Attribute{
+					"note": schema.StringAttribute{
+						Description: "Note for the secret rotation reminder",
+						Computed:    false,
+						Optional:    true,
```
maidul98, line 99: The secret_reminder object is optional but the params of the secret_reminder should be required i think

internal/provider/resource/secret_resource.go
```diff
@@ -219,14 +241,26 @@ func (r *secretResource) Create(ctx context.Context, req resource.CreateRequest,
...
 		plan.WorkspaceId = types.StringValue(serviceTokenDetails.Workspace)
 	} else if r.client.Config.IsMachineIdentityAuth {
+		
+		// null check secret reminder
+		var secretReminderNote string
+		var secretReminderRepeatDays int64
+
+		if plan.SecretReminder != nil {
+			secretReminderNote = plan.SecretReminder.Note.ValueString()
+			secretReminderRepeatDays = plan.SecretReminder.RepeatDays.ValueInt64()
```
maidul98, line 251: Was there a reason you didn't add `plan.SecretReminder.RepeatDays.ValueInt64()` directly to `CreateRawSecretsV3`?

internal/provider/resource/secret_resource.go
meetcshah19: The note is optional, the repetition frequency isn't. On the UI we show it as a reminder to rotate the secret by default.
Screenshot 2024-09-11 at 2 40 27 AM

internal/provider/resource/secret_resource.go
meetcshah19: Because the default value is 0 if null and the API accepts it, which shouldn't be allowed.

meetcshah19: We can also choose to handle this at the API level.

internal/provider/resource/secret_resource.go
meetcshah19: When using `SecretReminder SecretReminder`, I get this when a reminder isn't present in a secret
```
Error: Value Conversion Error
│ 
│   with infisical_secret.mongo_secret,
│   on resource.tf line 32, in resource "infisical_secret" "mongo_secret":
│   32: resource "infisical_secret" "mongo_secret" {
│ 
│ An unexpected error was encountered trying to build a value. This is always an error in the provider. Please report the following to
│ the provider developer:
│ 
│ Received null value, however the target type cannot handle null values. Use the corresponding `types` package type, a pointer type or a
│ custom type that handles null values.
│ 
│ Path: secret_reminder
│ Target Type: resource.SecretReminder
│ Suggested `types` Type: basetypes.ObjectValue
│ Suggested Pointer Type: *resource.SecretReminder

File: go.mod [hunk 5]
```diff
@@ -1,13 +1,15 @@
 module terraform-provider-infisical
 
-go 1.19
+go 1.21
```
maidul98, line 3: Why was the go version updated? This could be problematic

internal/provider/resource/secret_resource.go
maidul98: Because of our API design, i think it makes sense to check for this as null. Resolving

internal/provider/resource/secret_resource.go
maidul98: Right, resolving so we don't have to make API changes

maidul98 (approved):
Nice work!

PR #51, opened 2024-09-13
feat: added identity oidc auth resource

sheensantoscapadngan:
This PR will allow users to manage OIDC auth configurations of machine identities via Terraform


maidul98 (approved):
lgtm

PR #56, opened 2024-09-24
fix(project-user-resource): tf error if membership gets removed in infisical

DanielHougaard:
This PR resolves a bug, causing TF to error out with "Project membership not found", if the membership has been removed in Infisical. This can happen if the user manually leaves the project, or if they are removed through other means, such as by an admin.

We should always assume the terraform definitions should be the true state of Infisical. So this makes sure that if the project membership isn't found in Infisical, it will be re-created as per defined in Terraform, and then the TF state will be back in sync.

> [!CAUTION]
> This PR is dependent on https://github.com/Infisical/infisical/pull/2480, which should be merged and deployed before this is released.

### DIFF
--- a/internal/provider/resource/project_user_resource.go
+++ b/internal/provider/resource/project_user_resource.go
@@ -3,6 +3,7 @@ package resource
 import (
 	"context"
 	"fmt"
+	"strings"
 	infisical "terraform-provider-infisical/internal/client"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/provider/resource/project_user_resource.go
```diff
@@ -345,11 +346,19 @@ func (r *ProjectUserResource) Read(ctx context.Context, req resource.ReadRequest
 		ProjectID: state.ProjectID.ValueString(),
 		Username:  state.Username.ValueString(),
 	})
+
 	if err != nil {
+		// If the membership isn't found for whatever reason (user left the project, etc), remove the resource so we can re-create it, and fix the state.
+		if strings.Contains(err.Error(), "Project membership not found") {
```
maidul98, line 352: We should make use of the `ErrNotFound` error instead of this string based apprach

internal/provider/resource/project_user_resource.go
```diff
@@ -345,11 +346,19 @@ func (r *ProjectUserResource) Read(ctx context.Context, req resource.ReadRequest
 		ProjectID: state.ProjectID.ValueString(),
 		Username:  state.Username.ValueString(),
 	})
+
 	if err != nil {
+		// If the membership isn't found for whatever reason (user left the project, etc), remove the resource so we can re-create it, and fix the state.
```
maidul98, line 351: nit: comment is not needed as it's clear what the logic is doing

maidul98 (approved):

maidul98 (approved):

PR #59, opened 2024-10-02
feat: gcp integration support

DanielHougaard:
> [!CAUTION]
> This PR is dependent on https://github.com/Infisical/infisical/pull/2522

This PR introduces an entirely new resource type, which is integrations. I've built it with the thought that we should at least be able to reuse our API call functions. The API call functions would work to create/update/delete any type of integration.

Please see the resource.tf file for an example of how to use this. Currently we do not support Google OAuth, but instead we support service account auth _(must be in json format)_.


internal/provider/resource/integration_gcp_secret_manager.go
```diff
@@ -0,0 +1,380 @@
...
+
+// Metadata returns the resource type name.
+func (r *IntegrationGcpSecretManagerResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_integration_gcp_secret_manager"
+}
+
+// Schema defines the schema for the resource.
+func (r *IntegrationGcpSecretManagerResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Create project users & save to Infisical. Only Machine Identity authentication is supported for this data source",
```
sheensantoscapadngan, line 53: update

internal/provider/resource/integration_gcp_secret_manager.go
```diff
@@ -0,0 +1,380 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+)
+
+// Ensure the implementation satisfies the expected interfaces.
+var (
+	_ resource.Resource = &IntegrationGcpSecretManagerResource{}
+)
+
+// NewProjectResource is a helper function to simplify the provider implementation.
+func NewIntegrationResource() resource.Resource {
```
sheensantoscapadngan, line 22: we might need to add `Gcp` prefix here so that this resource is unique within the package. there will be conflict when we add more integrations

internal/provider/resource/integration_gcp_secret_manager.go
```diff
@@ -0,0 +1,380 @@
...
+	return &IntegrationGcpSecretManagerResource{}
+}
+
+// ProjectUserResource is the resource implementation.
```
sheensantoscapadngan, line 26: update comment

internal/provider/resource/integration_gcp_secret_manager.go
```diff
@@ -0,0 +1,380 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+}
+
+// Read refreshes the Terraform state with the latest data.
+func (r *IntegrationGcpSecretManagerResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	if !r.client.Config.IsMachineIdentityAuth {
+		resp.Diagnostics.AddError(
+			"Unable to read identity role",
```
sheensantoscapadngan, line 228: update

internal/provider/resource/integration_gcp_secret_manager.go
```diff
@@ -0,0 +1,380 @@
...
+			"Unable to get integration",
+			err.Error(),
+		)
+		return
+	}
+
+	state.GCPProjectID = types.StringValue(integration.Integration.AppID)
+	state.SecretPath = types.StringValue(integration.Integration.SecretPath)
+	state.EnvironmentID = types.StringValue(integration.Integration.EnvID)
+	state.IntegrationAuthID = types.StringValue(integration.Integration.IntegrationAuthID)
```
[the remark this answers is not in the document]
DanielHougaard: Automatically handled by Terraform

internal/client/integrations.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	if response.IsError() {
+		return GetIntegrationResponse{}, fmt.Errorf("CallGetIntegration: Unsuccessful response. [response=%s]", string(response.Body()))
+	}
+
+	return body, nil
+}
+
+func (client Client) CallUpdateIntegration(request UpdateIntegrationRequest) (UpdateIntegrationResponse, error) {
```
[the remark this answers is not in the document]
DanielHougaard: Sorry this is a pattern we have in the CLI and Operator, I forgot for a moment that we don't use it in the TF provider (but we probably should)

internal/client/integrations_auth.go
```diff
@@ -0,0 +1,53 @@
...
+		return DeleteIntegrationAuthResponse{}, fmt.Errorf("CallDeleteIntegrationAuth: Unable to complete api request [err=%s]", err)
+	}
+
+	if response.IsError() {
+		return DeleteIntegrationAuthResponse{}, fmt.Errorf("CallDeleteIntegrationAuth: Unsuccessful response. [response=%s]", string(response.Body()))
+	}
+
+	return body, nil
+
+}
```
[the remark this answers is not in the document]
DanielHougaard: I really wanted the integration auth and integration itself to be consolidated into a single resource, and when the resource is deleted, the auth is deleted with it.

sheensantoscapadngan (approved):

PR #61, opened 2024-10-04
feat(integrations): AWS & CircleCI Integrations

DanielHougaard:
This PR introduces integrations for both AWS Parameter Store, AWS Secrets Manager, and Circle CI.


internal/client/model.go
```diff
@@ -1470,7 +1472,7 @@ type CreateIntegrationRequest struct {
 	Region              string `json:"region,omitempty"`
 	Scope               string `json:"scope,omitempty"`
 
-	Metadata IntegrationMetadata `json:"metadata,omitempty"`
+	Metadata map[string]interface{} `json:"metadata,omitempty"`
```
sheensantoscapadngan, line 1475: why are we moving away from using the IntegrationMetadata type?

internal/provider/resource/integration_aws_parameter_store.go
```diff
@@ -0,0 +1,501 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+)
+
+// Ensure the implementation satisfies the expected interfaces.
+var (
+	_ resource.Resource = &IntegrationAWSParameterStoreResource{}
+)
+
+// NewProjectResource is a helper function to simplify the provider implementation.
```
sheensantoscapadngan, line 24: nit: update comment

internal/provider/resource/integration_aws_parameter_store.go
```diff
@@ -0,0 +1,501 @@
...
+	client *infisical.Client
+}
+
+type AwsParameterStoreMetadataStruct struct {
+	SecretAWSTag []struct {
+		Key   string `json:"key,omitempty"`
+		Value string `json:"value,omitempty"`
+	} `json:"secretAWSTag,omitempty"`
+	ShouldDisableDelete bool `json:"shouldDisableDelete,omitempty"`
+}
```
sheensantoscapadngan, line 40: oh I see. we make the metadata property in the request object generic but we add the specific metadata types in the integration resource?

internal/provider/resource/integration_aws_parameter_store.go
```diff
@@ -0,0 +1,501 @@
...
+	AwsTags             []awsTag `tfsdk:"aws_tags"`
+	ShouldDisableDelete bool     `tfsdk:"should_disable_delete"`
+}
+
+type awsTag struct {
+	Key   types.String `tfsdk:"key"`
+	Value types.String `tfsdk:"value"`
+}
+
+// projectResourceSourceModel describes the data source data model.
```
sheensantoscapadngan, line 52: update

internal/provider/resource/integration_aws_parameter_store.go
```diff
@@ -0,0 +1,501 @@
...
+		// Load AWS tags into metadata
+		for _, tag := range options.AwsTags {
+			metadata.SecretAWSTag = append(metadata.SecretAWSTag, struct {
+				Key   string `json:"key,omitempty"`
+				Value string `json:"value,omitempty"`
+			}{
+				Key:   tag.Key.ValueString(),
+				Value: tag.Value.ValueString(),
+			})
+		}
```
sheensantoscapadngan, line 88: to test: check what happens with reordering of tags

internal/provider/resource/integration_aws_parameter_store.go
```diff
@@ -0,0 +1,501 @@
...
+									PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
+								},
+								"value": schema.StringAttribute{
+									Description:   "The value of the tag.",
+									Optional:      true,
+									PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
+								},
+							},
+						},
+					},
```
sheensantoscapadngan, line 133: let's use a SetNestedAttribute here instead so that this property is order-agnostic?

internal/provider/resource/integration_aws_parameter_store.go
```diff
@@ -0,0 +1,501 @@
...
+		return
+	}
+}
+
+// Read refreshes the Terraform state with the latest data.
+func (r *IntegrationAWSParameterStoreResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+
+	if !r.client.Config.IsMachineIdentityAuth {
+		resp.Diagnostics.AddError(
+			"Unable to create integration",
```
sheensantoscapadngan, line 293: unable to read integration

internal/provider/resource/integration_aws_parameter_store.go
```diff
@@ -0,0 +1,501 @@
...
+
+	var planOptions AwsParameterStoreOptions
+
+	if !state.Options.IsNull() {
+		diags := state.Options.As(ctx, &planOptions, basetypes.ObjectAsOptions{})
+		resp.Diagnostics.Append(diags...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+	}
```
sheensantoscapadngan, line 331: to check: won't this result in the option property not being updated if for example the integration metadata changed outside of TF?

internal/provider/resource/integration_aws_parameter_store.go
```diff
@@ -0,0 +1,501 @@
...
+
+	updateNeeded := false
+
+	if integration.Integration.Metadata.ShouldDisableDelete != planOptions.ShouldDisableDelete {
+		planOptions.ShouldDisableDelete = integration.Integration.Metadata.ShouldDisableDelete
+		updateNeeded = true
+	}
```
sheensantoscapadngan, line 338: curious, why can't we just set the plan values directly to the one returned by the `integration` object?

internal/provider/resource/integration_aws_parameter_store.go
```diff
@@ -0,0 +1,501 @@
...
+		for _, tag := range integration.Integration.Metadata.SecretAWSTag {
+			newTags = append(newTags, awsTag{
+				Key:   types.StringValue(tag.Key),
+				Value: types.StringValue(tag.Value),
+			})
+		}
+
+		planOptions.AwsTags = newTags
+		updateNeeded = true
+	}
```
sheensantoscapadngan, line 365: I think this can be simplified if we used a SetNestedAttribute... by then we can just assign it directly to the planOption property

internal/provider/resource/integration_aws_parameter_store.go
```diff
@@ -0,0 +1,501 @@
...
+
+	if updateNeeded {
```
sheensantoscapadngan, line 367: why do we need this condition?

internal/provider/resource/integration_aws_parameter_store.go
```diff
@@ -0,0 +1,501 @@
...
+		}
+
+		// Set the new options in the state
+		state.Options = newOptions
+	}
+
+	// Set the state.Options.
+	state.SecretPath = types.StringValue(integration.Integration.SecretPath)
+	state.IntegrationAuthID = types.StringValue(integration.Integration.IntegrationAuthID)
+
```
sheensantoscapadngan, line 406: hmmm how about the env property?

internal/provider/resource/integration_aws_secrets_manager.go
```diff
@@ -0,0 +1,538 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+)
+
+// Ensure the implementation satisfies the expected interfaces.
+var (
+	_ resource.Resource = &IntegrationAWSSecretsManagerResource{}
+)
+
+// NewProjectResource is a helper function to simplify the provider implementation.
```
sheensantoscapadngan, line 25: nit: update

internal/provider/resource/integration_aws_secrets_manager.go
```diff
@@ -0,0 +1,538 @@
...
+									PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
+								},
+								"value": schema.StringAttribute{
+									Description:   "The value of the tag.",
+									Optional:      true,
+									PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
+								},
+							},
+						},
+					},
```
sheensantoscapadngan, line 140: let's update to use SetNestedAttribute as well

internal/provider/resource/integration_aws_secrets_manager.go
```diff
@@ -0,0 +1,538 @@
...
+				Description:   "The AWS access key ID.",
+				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
+			},
+
+			"secret_access_key": schema.StringAttribute{
+				Sensitive:     true,
+				Required:      true,
+				Description:   "The AWS secret access key.",
+				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
+			},
```
sheensantoscapadngan, line 174: why does updating these require a complete replace?

internal/provider/resource/integration_aws_secrets_manager.go
```diff
@@ -0,0 +1,538 @@
...
+				Computed:      true,
+				Description:   "The ID of the integration, used internally by Infisical.",
+				PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
+			},
+
+			"aws_region": schema.StringAttribute{
+				Required:      true,
+				Description:   "The AWS region to sync secrets to.",
+				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
+			},
```
sheensantoscapadngan, line 160: it would help if we added possible values in the description as well

internal/provider/resource/integration_aws_secrets_manager.go
```diff
@@ -0,0 +1,538 @@
...
+			"Invalid plan",
+			"secrets_manager_path is required when mapping_behavior is 'many-to-one'",
+		)
+		return
+	}
+
+	if plan.MappingBehavior.ValueString() == MAPPING_BEHAVIOR_ONE_TO_ONE && (!plan.AWSPath.IsNull() && plan.AWSPath.ValueString() != "") {
+		resp.Diagnostics.AddError(
+			"Invalid plan",
+			"secrets_manager_path is not required when mapping_behavior is 'one-to-one'",
```
sheensantoscapadngan, line 258: maybe "should not be used" would be more appropriate?

internal/provider/resource/integration_aws_secrets_manager.go
```diff
@@ -0,0 +1,538 @@
...
+		return
+	}
+}
+
+// Read refreshes the Terraform state with the latest data.
+func (r *IntegrationAWSSecretsManagerResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+
+	if !r.client.Config.IsMachineIdentityAuth {
+		resp.Diagnostics.AddError(
+			"Unable to create integration",
```
sheensantoscapadngan, line 330: unable to read integration

internal/provider/resource/integration_aws_secrets_manager.go
```diff
@@ -0,0 +1,538 @@
...
+		for _, planTag := range planOptions.AwsTags {
+			if tag.Key == planTag.Key.ValueString() && tag.Value == planTag.Value.ValueString() {
+				found = true
+				break
+			}
+		}
+		if !found {
+			break
+		}
+	}
```
sheensantoscapadngan, line 388: same question here

internal/provider/resource/integration_aws_secrets_manager.go
```diff
@@ -0,0 +1,538 @@
...
+		}
+
+		// Set the new options in the state
+		state.Options = newOptions
+	}
+
+	// Set the state.Options.
+	state.SecretPath = types.StringValue(integration.Integration.SecretPath)
+	state.IntegrationAuthID = types.StringValue(integration.Integration.IntegrationAuthID)
+
```
sheensantoscapadngan, line 443: should we add the environment property as well?

internal/provider/resource/integration_circleci.go
```diff
@@ -0,0 +1,307 @@
...
+				"Unable to get integration",
+				err.Error(),
+			)
+		}
+		return
+	}
+
+	state.SecretPath = types.StringValue(integration.Integration.SecretPath)
+	state.IntegrationAuthID = types.StringValue(integration.Integration.IntegrationAuthID)
+
```
sheensantoscapadngan, line 223: environment as well?

internal/provider/resource/integration_circleci.go
```diff
@@ -0,0 +1,307 @@
...
+				Description:   "The ID of the integration, used internally by Infisical.",
+				PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
+			},
+
+			"circleci_token": schema.StringAttribute{
+				Required:      true,
+				Sensitive:     true,
+				Description:   "Your personal CircleCI token to authenticate with.",
+				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
+			},
```
sheensantoscapadngan, line 72: do we need to trigger a replace when this is updated?

internal/provider/resource/integration_aws_parameter_store.go
DanielHougaard: Asked for clarification via. Slack

internal/provider/resource/integration_aws_secrets_manager.go
DanielHougaard: Because the project ID cannot be updated, since the user shouldn't be moving integrations out of the project it was initially created in. If they want to do that, a new integration should take it's place

DanielHougaard: Because it would change the auth itself, which we currently don't support updating

internal/provider/resource/integration_circleci.go
DanielHougaard: Yes, this is auth related

internal/provider/resource/integration_aws_parameter_store.go
DanielHougaard: Tested with SetNestedAttribute: Nothing happens

internal/provider/resource/integration_aws_parameter_store.go
DanielHougaard: Resolved

internal/provider/resource/integration_aws_secrets_manager.go
```diff
@@ -435,6 +448,13 @@ func (r *IntegrationAWSSecretsManagerResource) Update(ctx context.Context, req r
 		return
 	}
 
+	plan.Options = plan.Options
+	plan.SecretPath = types.StringValue(updatedIntegration.Integration.SecretPath)
+	plan.IntegrationAuthID = types.StringValue(updatedIntegration.Integration.IntegrationAuthID)
+	plan.Environment = types.StringValue(updatedIntegration.Integration.Environment.Slug)
+
+	diags = resp.State.Set(ctx, plan)
```
sheensantoscapadngan, line 455: we're doing this twice
image

sheensantoscapadngan (approved):

PR #63, opened 2024-10-09
fix: project role drift when no conditions are set

DanielHougaard:
As the title says, this PR resolves project roles having an infinite drift when no conditions in the permissions are set.

### DIFF
--- a/internal/provider/resource/project_role_resource.go
+++ b/internal/provider/resource/project_role_resource.go
@@ -301,7 +301,7 @@ func (r *projectRoleResource) Read(ctx context.Context, req resource.ReadRequest
 			}
 		}
 
-		conditions := &projectRoleResourcePermissionCondition{}
+		var conditions *projectRoleResourcePermissionCondition
 
 		if el["conditions"] == nil {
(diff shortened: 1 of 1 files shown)


internal/provider/resource/project_role_resource.go
```diff
@@ -301,13 +301,21 @@ func (r *projectRoleResource) Read(ctx context.Context, req resource.ReadRequest
...
+		} else {
+			conditions = &projectRoleResourcePermissionCondition{
 				Environment: types.StringValue(environment),
 				SecretPath:  types.StringValue(secretPath),
-			},
+			}
+		}
+
+		permissionPlan = append(permissionPlan, projectRoleResourcePermissions{
+			Action:     types.StringValue(action),
```
maidul98, line 316: I think the action is now an array according to our response schema. Can you double check?

DanielHougaard: Already accounted for in a previous step in the code, see line 248-250

maidul98 (approved):

PR #64, opened 2024-10-10
feat: slug validators

DanielHougaard:
Added slug validators to terraform to prevent slugs from drifting as they sometimes take on a different form during processing on the API.

### DIFF
--- a/internal/pkg/terraform/validators.go
+++ b/internal/pkg/terraform/validators.go
@@ -6,7 +6,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 )
 
-var LowercaseRegexValidator = stringvalidator.RegexMatches(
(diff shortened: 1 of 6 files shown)


internal/pkg/terraform/validators.go
```diff
@@ -0,0 +1,12 @@
+package terraform
+
+import (
+	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+)
+
+var LowercaseRegexValidator = stringvalidator.RegexMatches(
+	regexp.MustCompile(`^(([^a-zA-Z0-9])|([a-z0-9]))*$`),
```
maidul98, line 10: Slugs should alphanumeric not non alphanumeric. I think this will allow things like `@!` . Ref https://www.npmjs.com/package/slugify

maidul98 (approved):

PR #66, opened 2024-10-23
fix: address secret folder path issue during import

sheensantoscapadngan:
This PR will resolve import issue regarding secret folders

### DIFF
--- a/internal/provider/resource/project_secret_folder.go
+++ b/internal/provider/resource/project_secret_folder.go
@@ -3,6 +3,7 @@ package resource
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"strings"
 	infisical "terraform-provider-infisical/internal/client"
 	infisicalclient "terraform-provider-infisical/internal/client"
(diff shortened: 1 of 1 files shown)


internal/provider/resource/project_secret_folder.go
```diff
@@ -296,14 +296,20 @@ func (r *projectSecretFolderResource) ImportState(ctx context.Context, req resou
 	// Remove leading and trailing slashes.
 	trimmedPath := strings.Trim(folder.Folder.Path, "/")
 	var name string
+	parentFolderPath := "/"
 
 	if trimmedPath != "" {
 		// Split the path and get the last element as the name.
 		pathParts := strings.Split(trimmedPath, "/")
 		name = pathParts[len(pathParts)-1]
+		if len(pathParts) > 1 {
```
maidul98, line 305: i think we can use:

```
import "path/filepath"

parentPath := filepath.Dir(originalPath)
```

maidul98 (approved):

PR #68, opened 2024-11-01
feat: add permission v2 

sheensantoscapadngan:


internal/provider/resource/project_role_resource.go
```diff
@@ -92,9 +116,41 @@ func (r *projectRoleResource) Schema(_ context.Context, _ resource.SchemaRequest
...
+							Description: "Describe what actions an entity can take.",
+							Required:    true,
+						},
+						"subject": schema.StringAttribute{
+							Description: "Describe the entity the permission pertains to.",
+							Required:    true,
+						},
+						"inverted": schema.BoolAttribute{
+							Description: "Whether rule forbids. Set this to true if permission forbids.",
+							Required:    true,
```
maidul98, line 135: i'm thinking we don't have to require this and can default it to false. what do you think?

internal/provider/resource/project_role_resource.go
```diff
@@ -92,9 +116,41 @@ func (r *projectRoleResource) Schema(_ context.Context, _ resource.SchemaRequest
...
+						},
+					},
+				},
+			},
 			"permissions": schema.ListNestedAttribute{
-				Required:    true,
-				Description: "The permissions assigned to the project role",
+				Optional:           true,
+				DeprecationMessage: "Use permissions_v2 instead as it allows you to be more granular with access control",
+				Description:        "(DEPRECATED, USE permissions_v2) The permissions assigned to the project role",
```
maidul98, line 153: I think it will be really helpful to point them to a document where we tell them how they can move from permisions v1 to v2 in terraform. The reason is because you can't just paste in the v1 permissions into v2. So i think people will be like how do i make this work. also since during each apply they get a big warning, i think they would want to know how they can fix this properly. what do you think?

maidul98 (commented):
small comments

internal/pkg/strings/main.go
```diff
@@ -9,3 +12,18 @@ func StringSplitAndTrim(input string, separator string) []string {
...
 }
+
+func NormalizeJSON(input string) (string, error) {
+	var parsed interface{}
+	if err := json.Unmarshal([]byte(input), &parsed); err != nil {
+		return "", err
+	}
+
+	// Use canonical JSON encoding settings
+	normalized, err := json.Marshal(parsed)
```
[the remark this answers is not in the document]
sheensantoscapadngan: yes exactly! 

this will make it so that order does not matter

internal/provider/resource/project_role_resource.go
sheensantoscapadngan: the problem with this one is we can't set a default value for it without making it into a`computed` field... and we can't make it into a computed field because it's inside a `Set`

internal/provider/resource/project_role_resource.go
sheensantoscapadngan: agreed... added reference to the migration guide. 

It's part of the PR here:
https://github.com/Infisical/infisical/pull/2689

internal/provider/resource/project_role_resource.go
```diff
@@ -35,14 +38,22 @@ type projectRoleResource struct {
...
-	Slug        types.String                     `tfsdk:"slug"`
-	ProjectSlug types.String                     `tfsdk:"project_slug"`
-	ID          types.String                     `tfsdk:"id"`
-	Permissions []projectRoleResourcePermissions `tfsdk:"permissions"`
+	Name          types.String                   `tfsdk:"name"`
+	Description   types.String                   `tfsdk:"description"`
+	Slug          types.String                   `tfsdk:"slug"`
+	ProjectSlug   types.String                   `tfsdk:"project_slug"`
+	ID            types.String                   `tfsdk:"id"`
+	Permissions   types.List                     `tfsdk:"permissions"`
```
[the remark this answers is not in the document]
maidul98: oh good to know

maidul98 (approved):

PR #74, opened 2024-11-19
Added acceptance testcases

akhilmhdh:
This PR adds acceptance test to ensure terraform stability. This includes most of the important resources like secret, secret folders, identity, identity auth, project identity, permission etc.

More will be coming soon


internal/testacc/project_environment_resource_test.go
```diff
@@ -0,0 +1,99 @@
+package testAcc
```
DanielHougaard, line 1: This entire test file is commented out, is this intentional?

internal/testacc/secret_import_resource_test.go
```diff
@@ -0,0 +1,157 @@
+package testAcc
```
DanielHougaard, line 1: Same as above here

.github/workflows/test.yml
```diff
@@ -77,5 +79,9 @@ jobs:
       - run: go mod download
       - env:
           TF_ACC: "1"
-        run: go test -v -cover ./internal/provider/
+          INFISICAL_HOST: ${{ secrets.INFISICAL_HOST }}
+          INFISICAL_UNIVERSAL_AUTH_CLIENT_ID: ${{ secrets.INFISICAL_UNIVERSAL_AUTH_CLIENT_ID }}
+          INFISICAL_UNIVERSAL_AUTH_CLIENT_SECRET: ${{ secrets.INFISICAL_UNIVERSAL_AUTH_CLIENT_SECRET }}
+          INFISICAL_TEST_ORG_ID: ${{ secrets.INFISICAL_TEST_ORG_ID }}
+        run: go test ./... -v $(TESTARGS) -timeout 120m
```
DanielHougaard, line 86: All acceptance tests are failing due to `TESTARGS` command not found

.github/workflows/test.yml
```diff
@@ -77,5 +79,9 @@ jobs:
       - run: go mod download
       - env:
           TF_ACC: "1"
-        run: go test -v -cover ./internal/provider/
+          INFISICAL_HOST: ${{ secrets.INFISICAL_HOST }}
+          INFISICAL_UNIVERSAL_AUTH_CLIENT_ID: ${{ secrets.INFISICAL_UNIVERSAL_AUTH_CLIENT_ID }}
+          INFISICAL_UNIVERSAL_AUTH_CLIENT_SECRET: ${{ secrets.INFISICAL_UNIVERSAL_AUTH_CLIENT_SECRET }}
+          INFISICAL_TEST_ORG_ID: ${{ secrets.INFISICAL_TEST_ORG_ID }}
```
DanielHougaard, line 85: Testing locally:

Makefile: 
```
testacc:
	TF_ACC=1 go test ./... -v $(TESTARGS) -timeout 120m
```

`./...` results in no tests being run. If I replace `./...` with `./internal/testacc`, it works. Are you sure `./...` works?

varonix0 (changes requested):
After tweaking things slightly locally, I was able to get the tests to run, and it works smoothly after that! Good work, I left a few comments for you. 

We should defiantly make this a requirement for all future PR's, that the PR's also include covering tests.

internal/testacc/project_environment_resource_test.go
akhilmhdh: Yes for both environment and imports.

Both are the moment incompatiable to test with terraform. Need modification for both

internal/testacc/secret_import_resource_test.go
akhilmhdh: Answered above

.github/workflows/test.yml
akhilmhdh: Hmm let me check it out. The makefile has the same command.

PR #78, opened 2024-12-09
fix: resolved update not change the integration auth

akhilmhdh:
This PR fixes integration not updating when changing from auth fields and integration fields.

Before merging this, [this PR](https://github.com/Infisical/infisical/pull/2855) should be merged

### DIFF
--- a/internal/client/integrations_auth.go
+++ b/internal/client/integrations_auth.go
@@ -35,6 +35,26 @@ func (client Client) CreateIntegrationAuth(request CreateIntegrationAuthRequest)
 	return body, nil
 }
 
+func (client Client) UpdateIntegrationAuth(request UpdateIntegrationAuthRequest) (CreateIntegrationAuthResponse, error) {
+	var body CreateIntegrationAuthResponse
+	response, err := client.Config.HttpClient.
+		R().
+		SetResult(&body).
+		SetHeader("User-Agent", USER_AGENT).
+		SetBody(request).
+		Patch("api/v1/integration-auth/" + request.IntegrationAuthId)
+
+	if err != nil {
+		return CreateIntegrationAuthResponse{}, fmt.Errorf("UpdateIntegrationAuth: Unable to complete api request [err=%s]", err)
+	}
+
+	if response.IsError() {
+		return CreateIntegrationAuthResponse{}, fmt.Errorf("UpdateIntegrationAuth: Unsuccessful response. [response=%s]", string(response.Body()))
+	}
+
+	return body, nil
+}
+
 // Deleting integration auth triggers a cascade effect, that will also delete the associated integration.
 func (client Client) DeleteIntegrationAuth(request DeleteIntegrationAuthRequest) (DeleteIntegrationAuthResponse, error) {
 	var body DeleteIntegrationAuthResponse
--- a/internal/client/integrations_auth.go
+++ b/internal/client/integrations_auth.go
@@ -35,8 +35,8 @@ func (client Client) CreateIntegrationAuth(request CreateIntegrationAuthRequest)
 	return body, nil
 }
 
-func (client Client) UpdateIntegrationAuth(request UpdateIntegrationAuthRequest) (CreateIntegrationAuthResponse, error) {
-	var body CreateIntegrationAuthResponse
+func (client Client) UpdateIntegrationAuth(request UpdateIntegrationAuthRequest) (UpdateIntegrationAuthResponse, error) {
+	var body UpdateIntegrationAuthResponse
 	response, err := client.Config.HttpClient.
 		R().
 		SetResult(&body).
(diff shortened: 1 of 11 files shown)


internal/provider/resource/integration_aws_parameter_store.go
```diff
@@ -412,6 +410,26 @@ func (r *IntegrationAWSParameterStoreResource) Update(ctx context.Context, req r
...
+	}
+
+	_, err = r.client.UpdateIntegrationAuth(updateIntegrationAuthRequest)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error updating integration auth",
+			err.Error(),
+		)
+		return
+	}
```
sheensantoscapadngan, line 431: Better if we move this below the other validations so that we can somehow prevent inconsistency between TF state and integrationAuth state. Maybe we move it to directly before the ff:
image

internal/provider/resource/integration_aws_secrets_manager.go
```diff
@@ -444,14 +458,36 @@ func (r *IntegrationAWSSecretsManagerResource) Update(ctx context.Context, req r
...
-			return
-		}
+	_, err = r.client.UpdateIntegrationAuth(updateIntegrationAuthRequest)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error updating integration auth",
+			err.Error(),
+		)
+		return
+	}
```
sheensantoscapadngan, line 475: same for this one

internal/client/integrations_auth.go
```diff
@@ -35,6 +35,26 @@ func (client Client) CreateIntegrationAuth(request CreateIntegrationAuthRequest)
 	return body, nil
 }
 
+func (client Client) UpdateIntegrationAuth(request UpdateIntegrationAuthRequest) (CreateIntegrationAuthResponse, error) {
```
sheensantoscapadngan, line 38: let's have a separate `UpdateIntegrationAuthResponse` interface

internal/provider/resource/integration_aws_secrets_manager.go
```diff
@@ -444,14 +458,36 @@ func (r *IntegrationAWSSecretsManagerResource) Update(ctx context.Context, req r
 		return
 	}
 
-	var planOptions AwsSecretsManagerOptions
+	if authMethod == pkg.AwsAuthMethodAccessKey {
+		updateIntegrationAuthRequest.AccessId = plan.AccessKeyID.ValueString()
+		updateIntegrationAuthRequest.AccessToken = plan.SecretAccessKey.ValueString()
+	} else if authMethod == pkg.AwsAuthMethodAssumeRole {
+		updateIntegrationAuthRequest.AWSAssumeIamRoleArn = plan.AssumeRoleArn.ValueString()
+	}
```
sheensantoscapadngan, line 467: I noticed that we already have a `RequiresReplace` modifier for these fields. Would this make any difference? Same is the case for the other integration resources 
image

akhilmhdh: Need to connect with @DanielHougaard over this one.

internal/provider/resource/integration_aws_parameter_store.go
```diff
@@ -412,6 +405,26 @@ func (r *IntegrationAWSParameterStoreResource) Update(ctx context.Context, req r
 		}
 	}
 
+	updateIntegrationAuthRequest := infisical.UpdateIntegrationAuthRequest{
+		Integration:       infisical.IntegrationAuthTypeAwsSecretsManager,
+		IntegrationAuthId: plan.IntegrationAuthID.String(),
```
sheensantoscapadngan, line 410: I think for this one you have to use `.ValueString()` because `String()` returns a value not intended for handling. For instance in my testing, the string value returned by `String()` has "" in it:

image
image

internal/provider/resource/integration_circleci.go
```diff
@@ -256,11 +253,27 @@ func (r *IntegrationCircleCIResource) Update(ctx context.Context, req resource.U
 		return
 	}
 
+	_, err := r.client.UpdateIntegrationAuth(infisical.UpdateIntegrationAuthRequest{
+		Integration:       infisical.IntegrationAuthTypeCircleCi,
+		IntegrationAuthId: plan.IntegrationAuthID.String(),
```
sheensantoscapadngan, line 258: same here

internal/provider/resource/integration_databricks.go
```diff
@@ -253,11 +250,26 @@ func (r *IntegrationDatabricksResource) Update(ctx context.Context, req resource
 		return
 	}
 
+	_, err := r.client.UpdateIntegrationAuth(infisical.UpdateIntegrationAuthRequest{
+		Integration:       infisical.IntegrationAuthTypeDatabricks,
+		IntegrationAuthId: plan.IntegrationAuthID.String(),
```
sheensantoscapadngan, line 255: and here

internal/provider/resource/integration_aws_parameter_store.go
```diff
@@ -428,6 +441,8 @@ func (r *IntegrationAWSParameterStoreResource) Update(ctx context.Context, req r
 		Metadata:    metadataMap,
 		Environment: plan.Environment.ValueString(),
 		SecretPath:  plan.SecretPath.ValueString(),
+		Region:      plan.AWSRegion.ValueString(),
+		Path:        plan.AWSPath.String(),
```
sheensantoscapadngan, line 445: and here too

internal/provider/resource/integration_aws_secrets_manager.go
```diff
@@ -434,8 +429,38 @@ func (r *IntegrationAWSSecretsManagerResource) Update(ctx context.Context, req r
...
+		resp.Diagnostics.AddError(
+			"Invalid plan",
+			"secrets_manager_path should not be used when mapping_behavior is 'one-to-one'",
+		)
+		return
+	}
+
+	updateIntegrationAuthRequest := infisical.UpdateIntegrationAuthRequest{
+		Integration:       infisical.IntegrationAuthTypeAwsSecretsManager,
+		IntegrationAuthId: plan.IntegrationAuthID.String(),
```
sheensantoscapadngan, line 460: and here

sheensantoscapadngan (approved):

PR #81, opened 2024-12-11
fix(oidc-auth): comma formatted bound claims required space formatting

DanielHougaard:
Fixed bug causing Terraform to provide invalid plan error when `bound_claims.repository` was not formatted WITH spaces between comma-separated items.

The OIDC auth resource now supports both with/without spaces between comma-separated items.

### DIFF
--- a/internal/pkg/modifiers/comma_space_map_modifier.go
+++ b/internal/pkg/modifiers/comma_space_map_modifier.go
@@ -1,0 +1,60 @@
+package pkg
+
+import (
+	"context"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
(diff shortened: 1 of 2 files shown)


internal/pkg/modifiers/comma_space_map_modifier.go
```diff
@@ -0,0 +1,81 @@
...
+	if !req.ConfigValue.IsNull() {
+		configElements := req.ConfigValue.Elements()
+		// Look at first value to determine format
+		for _, v := range configElements {
+			if str, ok := v.(types.String); ok && !str.IsNull() {
+				configFormat = strings.Contains(str.ValueString(), ", ")
+				break
+			}
+		}
+	}
```
sheensantoscapadngan, line 41: why do we need this btw? I think an alternative would be to just always modify the map values to be ", " separated

sheensantoscapadngan (approved):

PR #85, opened 2025-01-08
fix: addressed issue where integration is disabled when updated

sheensantoscapadngan:
This ensures that when integrations are updated, it doesn't disable the integration sync entirely.

### DIFF
--- a/internal/provider/resource/integration_circleci.go
+++ b/internal/provider/resource/integration_circleci.go
@@ -274,6 +274,7 @@ func (r *IntegrationCircleCIResource) Update(ctx context.Context, req resource.U
 		App:         plan.CircleCIProjectID.ValueString(), // Needs to be the project slug
 		AppID:       plan.CircleCIProjectID.ValueString(), // Needs to be the project ID
 		Owner:       plan.CircleCIOrgSlug.ValueString(),   // Needs to be the organization slug
+		IsActive:    true,
 	})
 
 	if err != nil {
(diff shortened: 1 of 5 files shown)


internal/provider/resource/integration_circleci.go
```diff
@@ -274,6 +274,7 @@ func (r *IntegrationCircleCIResource) Update(ctx context.Context, req resource.U
 		App:         plan.CircleCIProjectID.ValueString(), // Needs to be the project slug
 		AppID:       plan.CircleCIProjectID.ValueString(), // Needs to be the project ID
 		Owner:       plan.CircleCIOrgSlug.ValueString(),   // Needs to be the organization slug
+		IsActive:    true,
```
DanielHougaard, line 277: Is this necessary now that we're doing `omitempty` on `isActive`? If we're omitting on empty, it should be undefined on server-side, which means it won't be modified on the API right?

internal/provider/resource/integration_databricks.go
```diff
@@ -270,6 +270,7 @@ func (r *IntegrationDatabricksResource) Update(ctx context.Context, req resource
 		Environment: plan.Environment.ValueString(),
 		SecretPath:  plan.SecretPath.ValueString(),
 		App:         plan.DatabricksSecretScope.ValueString(),
+		IsActive:    true,
```
DanielHougaard, line 273: Same as above

DanielHougaard (changes requested):

internal/provider/resource/integration_circleci.go
sheensantoscapadngan: Yeah we do this so that the Infisical terraform provider doesn't break when it's updated to the latest version when the API is left as is. 

Note that in the other PR, I had to make the `isActive` field optional
image

DanielHougaard (approved):

PR #87, opened 2025-01-12
show how to use the JWT for OIDC auth in the docs

lawndoc:
It wasn't immediately clear how to use the OIDC authentication method from the [provider documentation](https://registry.terraform.io/providers/Infisical/infisical/latest/docs#nested-schema-for-authoidc). Currently, you need to read the source code to uncover the environment variable that is needed.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -53,7 +53,7 @@ Optional:
 
 Optional:
 
-- `identity_id` (String, Sensitive) Machine identity ID. Used to fetch/modify secrets for a given project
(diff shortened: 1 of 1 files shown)


lawndoc:
References:

https://github.com/Infisical/terraform-provider-infisical/blob/342fb1ebe7a0ff864db3d7e75b5536a484660f1e/internal/client/login.go#L50-L51

https://github.com/Infisical/terraform-provider-infisical/blob/342fb1ebe7a0ff864db3d7e75b5536a484660f1e/internal/client/constants.go#L12

DanielHougaard:
Hi @lawndoc, you can't update the docs directly because they're auto generated. Please use `go generate ./...` to generate the docs, and make the appropriate changes to the go codebase itself.

PR #93, opened 2025-01-29
docs: provider environment variables

DanielHougaard:
This PR adds documentation for all supported provider environment variables except service token. I also slightly updated the description for the `client_id` and `client_secret` field, as they are considered legacy, and the `auth` attribute should be used instead.


maidul98 (approved):
lgtm, we should add example of the provider being set up with the `auth` field in the other PR you have open. right now, we don't have a good example of how to use the auth feild

PR #100, opened 2025-02-18
Allow loading in oidc token

maidul98:
Needed to load in token in Terraform cloud


internal/client/login.go
```diff
@@ -48,14 +48,19 @@ func (client Client) GetServiceTokenDetailsV2() (GetServiceTokenDetailsResponse,
 }
 
 func (client Client) OidcMachineIdentityAuth() (string, error) {
-	authJwt := os.Getenv(INFISICAL_AUTH_JWT_NAME)
+	tokenEnvironmentName := client.Config.OidcTokenEnvName
+	if tokenEnvironmentName == "" {
+		tokenEnvironmentName = "TFC_WORKLOAD_IDENTITY_TOKEN" // default
+	}
+
+	authJwt := os.Getenv(tokenEnvironmentName)
```
sheensantoscapadngan, line 56: I think for this not to be breaking for existing pipelines, we should try and read the INFISICAL_AUTH_JWT_NAME env first and only resort to the `OidcTokenEnvName` if no value is found there

sheensantoscapadngan (approved):

PR #102, opened 2025-03-05
feat: secret imports

DanielHougaard:
This PR adds support for secret imports. The PR is dependent on https://github.com/Infisical/infisical/pull/3181, because we currently don't support fetching secrets by ID on our API.


internal/client/secrets.go
```diff
@@ -119,6 +119,25 @@ func (client Client) GetSingleSecretByNameV3(request GetSingleSecretByNameV3Requ
...
+		SetHeader("User-Agent", USER_AGENT).
+		Get(fmt.Sprintf("api/v3/secrets/raw/id/%s", request.ID))
+
+	if err != nil {
+		return GetSingleSecretByIDV3Response{}, fmt.Errorf("CallGetSingleSecretByIDV3: Unable to complete api request [err=%s]", err)
+	}
+
+	if response.IsError() {
+		return GetSingleSecretByIDV3Response{}, fmt.Errorf("CallGetSingleSecretByIDV3: Unsuccessful response. Please make sure your secret path, workspace and environment name are all correct [response=%s]", response)
+	}
```
sheensantoscapadngan, line 136: the only thing that would matter is the secret ID right?

internal/provider/resource/secret_resource.go
```diff
@@ -692,3 +693,43 @@ func (r *secretResource) Delete(ctx context.Context, req resource.DeleteRequest,
 	}
 
 }
+
+func (r *secretResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+
+	secret, err := r.client.GetSingleSecretByIDV3(infisical.GetSingleSecretByIDV3Request{
+		ID: req.ID,
```
sheensantoscapadngan, line 700: from a use-case POV, where do I get the secret ID? it's currently not obvious in our UI, so we might have to add something there?

internal/provider/resource/secret_resource.go
```diff
@@ -692,3 +693,43 @@ func (r *secretResource) Delete(ctx context.Context, req resource.DeleteRequest,
...
+	})
+
+	if err != nil {
+		if err == infisical.ErrNotFound {
+			resp.Diagnostics.AddError(
+				"Secret not found",
+				"The secret with the given ID was not found",
+			)
+		} else {
```
sheensantoscapadngan, line 709: looks like we're missing this 

	if response.StatusCode() == http.StatusNotFound {
		return {}, ErrNotFound
	}

internal/provider/resource/secret_resource.go
DanielHougaard: Resolved after convo on Slack. Keeping as-is

sheensantoscapadngan (approved):

PR #104, opened 2025-03-11
feat(secret-sync): aws parameter store & secrets manager syncs

DanielHougaard:
This PR adds support for AWS Secrets Manager and AWS Parameter store. Additionally, the base secret syncs now support Sync Options. This was a needed expansion in order to allow for customization of the AWS secret syncs.


internal/provider/resource/secret_sync/secret_sync_aws_secrets_manager.go
```diff
@@ -0,0 +1,409 @@
...
+			} else {
+				syncOptionsMap["aws_kms_key_id"] = types.StringNull() // Add a null value for missing attributes
+			}
+
+			if secretSync.SyncOptions["syncSecretMetadataAsTags"] != nil {
+
+				syncSecretMetadataAsTags, ok := secretSync.SyncOptions["syncSecretMetadataAsTags"].(bool)
+				if !ok {
+					syncSecretMetadataAsTags = false
+				}
```
sheensantoscapadngan, line 157: shouldn't we be throwing an error for this because we got a different type?

sheensantoscapadngan (approved):

PR #106, opened 2025-03-14
feat(project-resource): template support

DanielHougaard:
This PR adds templating support for projects created via. Terraform. It's only supported on the initial create, since there's no way to change the template once a project is created. We use a RequireReplace modifier to insure that the template isn't changed.

Additionally, I added a RequireReplace on the project_user resource's projectId field. We should be replacing memberships if the project ID changes (such as if the project gets replaced)

### DIFF
--- a/internal/provider/resource/project_resource.go
+++ b/internal/provider/resource/project_resource.go
@@ -34,11 +34,12 @@ type projectResource struct {
 
 // projectResourceSourceModel describes the data source data model.
 type projectResourceModel struct {
-	Slug        types.String `tfsdk:"slug"`
-	ID          types.String `tfsdk:"id"`
-	Name        types.String `tfsdk:"name"`
-	Description types.String `tfsdk:"description"`
-	LastUpdated types.String `tfsdk:"last_updated"`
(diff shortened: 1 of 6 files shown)


internal/provider/resource/project_resource.go
```diff
@@ -66,6 +67,12 @@ func (r *projectResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 				Description: "The description of the project",
 				Optional:    true,
 			},
+			"template_name": schema.StringAttribute{
+				Description:   "The name of the template to use for the project",
+				Optional:      true,
+				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
+			},
```
sheensantoscapadngan, line 74: I worry that this `RequiresReplace` modifier might cause projects to be unintentionally deleted which would be super breaking (because project IDs used in workflows are hard-coded). Thoughts on just throwing an error here on update?

DanielHougaard: You're likely right, I changed it

sheensantoscapadngan (approved):

PR #110, opened 2025-03-25
feat(oidc-auth): claim metadata mapping

DanielHougaard:
This PR adds support for OIDC auth claim metadata mapping, which was recently introduced to OIDC auth.

### DIFF
--- a/internal/provider/resource/identity_oidc_auth.go
+++ b/internal/provider/resource/identity_oidc_auth.go
@@ -101,7 +101,7 @@ func (r *IdentityOidcAuthResource) Schema(_ context.Context, _ resource.SchemaRe
 			},
 
 			"claim_metadata_mapping": schema.MapAttribute{
-				Description: "The attributes that should be present in the permission metadata from the JWT.",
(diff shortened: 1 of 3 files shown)


internal/provider/resource/identity_oidc_auth.go
```diff
@@ -98,6 +99,16 @@ func (r *IdentityOidcAuthResource) Schema(_ context.Context, _ resource.SchemaRe
...
+
+			"claim_metadata_mapping": schema.MapAttribute{
+				Description: "The attributes that should be present in the permission metadata from the JWT.",
+				Optional:    true,
+				Computed:    true,
+				ElementType: types.StringType,
+				PlanModifiers: []planmodifier.Map{
+					mapplanmodifier.UseStateForUnknown(),
+				},
+			},
```
sheensantoscapadngan, line 111: it's not just about checking whether its present right? it's also about actually mapping the oidc JWT claims to the metadata? 

At first glance, it wasn't clear to me what this field is really for

PR #141, opened 2025-06-13
Delete protection for machine identity

x032205:


DanielHougaard (changes requested):
Breaks on apply, tested directly against prod test org

CleanShot 2025-06-16 at 22 49 22

internal/provider/resource/identity.go
```diff
@@ -53,6 +54,11 @@ func (r *IdentityResource) Schema(_ context.Context, _ resource.SchemaRequest, r
 				Description: "The name for the identity",
 				Required:    true,
 			},
+			"has_delete_protection": schema.BoolAttribute{
+				Description: "Whether the identity has delete protection, defaults to false",
+				Optional:    true,
+				Computed:    true,
```
sheensantoscapadngan, line 60: why is this marked as computed? wouldn't just `Optional:true` do the job here?

PR #142, opened 2025-06-18
AWS IAM dynamic secret

x032205:
All descriptions are based on docs. Schema is based on router schema.


examples/resources/infisical_dynamic_secret_aws_iam/resource.tf
```diff
@@ -0,0 +1,80 @@
...
+          {
+            "Effect": "Allow",
+            "Action": "s3:ListBucket",
+            "Resource": "*"
+          }
+        ]
+      }
+      EOT
+      user_groups                    = "group-a,group-b"
+      policy_arns                    = "arn:aws:iam::aws:policy/ReadOnlyAccess,arn:aws:iam::123456789012:policy/SpecificPolicy"
```
sheensantoscapadngan, line 52: why not have this outside the access_key_config? These are shared by both access_key and assume_role config

internal/provider/resource/dynamic_secret/dynamic_secret_aws_iam.go
```diff
@@ -0,0 +1,431 @@
...
+					"user_groups": schema.StringAttribute{
+						Optional:    true,
+						Description: "The AWS IAM groups that should be assigned to the created users. Multiple values can be provided by separating them with commas",
+					},
+					"policy_arns": schema.StringAttribute{
+						Optional:    true,
+						Description: "The AWS IAM managed policies that should be attached to the created users. Multiple values can be provided by separating them with commas",
+					},
+				},
+			},
```
sheensantoscapadngan, line 132: let's move this out and keep it at the same level as "method"

sheensantoscapadngan (approved):

PR #144, opened 2025-06-24
feat: k8s dynamic secret

x032205:


internal/provider/resource/dynamic_secret/dynamic_secret_kubernetes.go
```diff
@@ -0,0 +1,424 @@
...
+
+			var audiences []string
+			for i, v := range audiencesRaw {
+				s, ok := v.(string)
+				if !ok {
+					diags.AddError(
+						"Invalid audience element type",
+						"Expected audience at index "+string(rune(i))+" to be a string but got something else.",
+					)
+					continue
```
sheensantoscapadngan, line 377: why is this "continue" and not return?

sheensantoscapadngan:
there's an infinite drift that happens when audiences is set to `[]` in the config
image

internal/provider/resource/dynamic_secret/dynamic_secret_kubernetes.go
```diff
@@ -0,0 +1,424 @@
...
+						configurationMap["ca"] = configuration.ApiConfig.Ca.ValueString()
+					}
+				} else {
+					diags.AddError(
+						"Invalid Configuration",
+						"When auth_method is 'api', the 'api_config' block must be provided.",
+					)
+					return nil, diags
+				}
+			case "gateway":
```
sheensantoscapadngan, line 165: let's add proper handling for this one please!
image

internal/provider/resource/dynamic_secret/dynamic_secret_kubernetes.go
```diff
@@ -372,23 +376,29 @@ func NewDynamicSecretKubernetesResource() resource.Resource {
...
+					}
+					audiences = append(audiences, s)
 				}
-				audiences = append(audiences, s)
-			}
 
-			audiencesList, listDiags := types.ListValueFrom(ctx, types.StringType, audiences)
-			diags.Append(listDiags...)
+				audiencesList, listDiags := types.ListValueFrom(ctx, types.StringType, audiences)
+				diags.Append(listDiags...)
```
sheensantoscapadngan, line 398: I guess we're also missing a check for diagnostic errors here.. Can you please do it for the entire file?

sheensantoscapadngan (approved):

PR #149, opened 2025-06-28
feat: expand secret references for ephemeral resources

### LINKED ISSUE #150: Secret reference issue with ephemeral object
### **Issue Description**
The Infisical Terraform provider's `ephemeral` block fails to retrieve secrets when the target environment uses variable references (like `${common.AWS_ACCESS_KEY_ID}`) instead of direct secret values.

### **Steps to Reproduce**

1. **Working Configuration:**
   ```hcl
   ephemeral "infisical_secret" "aws_access_key_id" {
     name         = "AWS_ACCESS_KEY_ID"
     env_slug     = "infra"
     workspace_id = data.infisical_projects.platform.id
     folder_path  = "/"
   }
   ```
   Where `k8s-infra` environment has `AWS_ACCESS_KEY_ID` set to a direct value (e.g., `foobarkey...`)

2. **Failing Configuration:**
   Same Terraform code, but where `k8s-infra` environment has `AWS_ACCESS_KEY_ID` set to a variable reference (e.g., `${common.AWS_ACCESS_KEY_ID}`)

### **Infisical Project Structure**
```
Project: test-project
├── Environment: common
│   └── AWS_ACCESS_KEY_ID: "foobarkey..." (direct value)
└── Environment: infra
    └── AWS_ACCESS_KEY_ID: "${common.AWS_ACCESS_KEY_ID}" (variable reference)
```

### **Expected Behavior**
The `ephemeral` block should resolve the variable reference and retrieve the actual secret value from the referenced environment (`common`).

### **Actual Behavior**
- **Error:** `InvalidClientTokenId: The security token included in the request is invalid`
- **Root Cause:** The provider retrieves the literal string `${common.AWS_ACCESS_KEY_ID}` instead of resolving it to the actual AWS access key value

### **Workaround**
Manually copy the secret value from the source environment (`common`) to the target environment (`infra`) instead of using variable references.

### **Environment Details**
- **Provider:** `infisical/infisical` version `~> 0.15`
- **Terraform Version:** `>= 1.12`

jdmcmahan:
Fixes #150 

This PR builds on #39 by automatically expanding secret references for ephemeral secrets.

Currently, ephemeral secrets simply return the path of the reference instead of expanding to the referenced value. If this is the expected behavior, I can take a crack at updating the documentation instead.

Otherwise, the main change here is the addition of [`?expandSecretReferences`](https://infisical.com/docs/api-reference/endpoints/secrets/read#parameter-expand-secret-references) to `GetSingleRawSecretByNameV3`. This is defaulted to `true` for ephemeral resources.


internal/provider/resource/secret_resource.go
```diff
@@ -447,11 +447,12 @@ func (r *secretResource) Read(ctx context.Context, req resource.ReadRequest, res
...
-			Type:        "shared",
-			WorkspaceId: state.WorkspaceId.ValueString(),
-			Environment: state.EnvSlug.ValueString(),
-			SecretPath:  state.FolderPath.ValueString(),
+			SecretName:             state.Name.ValueString(),
+			Type:                   "shared",
+			WorkspaceId:            state.WorkspaceId.ValueString(),
+			Environment:            state.EnvSlug.ValueString(),
+			SecretPath:             state.FolderPath.ValueString(),
+			ExpandSecretReferences: false,
```
jdmcmahan, line 455: My (naïve) fear with defaulting to `true` here is that using a reference in a resource block might result in an inconsistent state after `apply` if the state switches to the resolved value:

```terraform
resource "infisical_secret" "blah" {
  name         = "SOME_SECRET"
  value        = "${prod.blah.value}" # this value would be different after apply
}
```

Jihad:
Just faced this today; heavily using deeply nested referenced values in my setup.. had multiple errors i couldn't figure out. 

Currently there's no other way to get ephemeral expanded secrets with this provider?

jdmcmahan:
> Just faced this today; heavily using deeply nested referenced values in my setup.. had multiple errors i couldn't figure out. 
> 
> Currently there's no other way to get ephemeral expanded secrets with this provider?

Not to my knowledge, but I'm just a fellow user. In the meantime I simply haven't been able to use ephemeral secrets, which is a massive problem because some providers require them.

I just reread the CONTRIBUTING guide and don't think I'm missing anything in this PR. This has been open for over a year now. Can a maintainer chime in on what's holding this up?

PR #151, opened 2025-07-02
feat(secret-syncs): 1password sync & app connection

DanielHougaard:
This PR adds support for the 1Password secret sync and app connection.


internal/provider/resource/app_connection/app_connection_1password.go
```diff
@@ -0,0 +1,141 @@
...
+
+func NewAppConnection1PasswordResource() resource.Resource {
+	return &AppConnectionBaseResource{
+		App:               infisical.AppConnectionApp1Password,
+		AppConnectionName: "1Password",
+		ResourceTypeName:  "_app_connection_1password",
+		AllowedMethods:    []string{AppConnection1PasswordAuthMethodApiToken},
+		CredentialsAttributes: map[string]schema.Attribute{
+			"instance_url": schema.StringAttribute{
+				Optional:    true,
```
sheensantoscapadngan, line 31: this should be required though right?

internal/provider/resource/app_connection/app_connection_1password.go
```diff
@@ -0,0 +1,141 @@
...
+				Description: "The URL of the 1Password Connect instance to connect to. For more details, refer to the documentation here infisical.com/docs/integrations/app-connections/1password",
+				Sensitive:   true,
+			},
+			"api_token": schema.StringAttribute{
+				Optional:    true,
+				Description: "The API token to use for authentication. For more details, refer to the documentation here infisical.com/docs/integrations/app-connections/1password",
+				Sensitive:   true,
```
sheensantoscapadngan, line 38: same here

image

internal/provider/resource/app_connection/app_connection_1password.go
```diff
@@ -0,0 +1,141 @@
...
+				return nil, diags
+			}
+			if credentials.ApiToken.IsNull() || credentials.ApiToken.ValueString() == "" {
+				diags.AddError(
+					"Unable to create 1Password app connection",
+					"api_token field must be defined in api-token method",
+				)
+				return nil, diags
+			}
+
```
sheensantoscapadngan, line 72: no need for these ones if we were to mark it as required in the property-level

PR #157, opened 2025-07-24
Updated policy requests to support multi-environment parameters instead of a single environment

carlosmonastyrski:
Related to https://github.com/Infisical/infisical/pull/4220

### DIFF
--- a/internal/provider/resource/access_approval_policy.go
+++ b/internal/provider/resource/access_approval_policy.go
@@ -316,7 +316,7 @@ func (r *accessApprovalPolicyResource) Read(ctx context.Context, req resource.Re
 
 	state.Approvers = approvers
 
-	if len(accessApprovalPolicy.AccessApprovalPolicy.Environments) > 0 {
+	if len(accessApprovalPolicy.AccessApprovalPolicy.Environments) > 0 && !state.EnvironmentSlugs.IsNull() {
 		// Extract environment slugs from the environment objects
 		var environmentSlugs []string
 		for _, env := range accessApprovalPolicy.AccessApprovalPolicy.Environments {
--- a/internal/provider/resource/access_approval_policy.go
+++ b/internal/provider/resource/access_approval_policy.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	infisical "terraform-provider-infisical/internal/client"
+	infisicaltf "terraform-provider-infisical/internal/pkg/terraform"
 
 	"github.com/hashicorp/terraform-plugin-framework/resource"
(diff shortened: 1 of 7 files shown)


internal/provider/resource/access_approval_policy.go
```diff
@@ -66,9 +67,10 @@ func (r *accessApprovalPolicyResource) Schema(_ context.Context, _ resource.Sche
 				Computed:      true,
 				PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
 			},
-			"environment_slug": schema.StringAttribute{
+			"environment_slugs": schema.ListAttribute{
 				Description: "The environment to apply the access approval policy to",
```
sheensantoscapadngan, line 71: let's update the description to be plural, so "environments"

internal/provider/resource/secret_approval_policy.go
```diff
@@ -68,9 +69,10 @@ func (r *secretApprovalPolicyResource) Schema(_ context.Context, _ resource.Sche
 				Computed:      true,
 				PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
 			},
-			"environment_slug": schema.StringAttribute{
+			"environment_slugs": schema.ListAttribute{
 				Description: "The environment to apply the secret approval policy to",
 				Required:    true,
+				ElementType: types.StringType,
```
sheensantoscapadngan, line 75: same set of comments for this resource here

internal/provider/resource/access_approval_policy.go
```diff
@@ -200,11 +202,12 @@ func (r *accessApprovalPolicyResource) Create(ctx context.Context, req resource.
 			Type: el.Type.ValueString(),
 		})
 	}
+	environments := infisicaltf.StringListToGoStringSlice(ctx, resp.Diagnostics, plan.EnvironmentSlugs)
 
 	accessApprovalPolicy, err := r.client.CreateAccessApprovalPolicy(infisical.CreateAccessApprovalPolicyRequest{
 		Name:              plan.Name.ValueString(),
 		ProjectSlug:       projectDetail.Slug,
-		Environment:       plan.EnvironmentSlug.ValueString(),
+		Environments:      environments,
```
sheensantoscapadngan, line 210: I think we should keep "Environment" there just in case the Infisical API server they are running on is not yet updated to the latest version

sheensantoscapadngan: review the other parts as well so that it works even when the API they've connected to is outdated

PR #158, opened 2025-07-24
feat: projects resource

sidwebworks:
This PR adds support for managing project templates in Infisical as a terraform resource.


internal/client/model.go
```diff
@@ -2341,3 +2341,77 @@ type DeleteSecretRotationRequest struct {
...
+type Role struct {
+	Name        string       `json:"name"`
+	Slug        string       `json:"slug"`
+	Permissions []Permission `json:"permissions"`
+}
+
+type Permission struct {
+	Subject   string               `json:"subject"`
+	Action    []string             `json:"action"`
+	Condition *PermissionCondition `json:"conditions,omitempty"`
```
sheensantoscapadngan, line 2378: Condition here should be a generic `map[string]any` right?

image

internal/provider/resource/project_template_resource.go
```diff
@@ -0,0 +1,676 @@
...
+			"name": schema.StringAttribute{
+				Description: "The name of the project template",
+				Required:    true,
+			},
+			"description": schema.StringAttribute{
+				Description: "The description of the project template",
+				Optional:    true,
+			},
+			"type": schema.StringAttribute{
+				Description: "The type of the project template",
```
sheensantoscapadngan, line 89: what are the possible values?

let's ensure that the doc is as complete as possible to other users

internal/provider/resource/project_template_resource.go
```diff
@@ -0,0 +1,676 @@
...
+							Description: "The name of the role",
+							Required:    true,
+						},
+						"slug": schema.StringAttribute{
+							Description: "The slug of the role",
+							Required:    true,
+						},
+						"permissions": schema.ListNestedAttribute{
+							Optional:    true,
+							Description: "The permissions assigned to the project identity specific privilege. Refer to the documentation here https://infisical.com/docs/internals/permissions for its usage.",
```
sheensantoscapadngan, line 108: the Description value here seems incorrect

internal/provider/resource/project_template_resource.go
```diff
@@ -0,0 +1,676 @@
...
+			Position: position.ValueInt64(),
+		}
+
+		envs = append(envs, env)
+	}
+
+	return envs, diags
+}
+
+func isDefaultPermissionScope(slug string) bool {
```
sheensantoscapadngan, line 674: I think this should be more like `isDefaultRole` right?

internal/provider/resource/project_template_resource.go
```diff
@@ -0,0 +1,699 @@
...
+		}
+
+		resp.Diagnostics.AddError(
+			"Error deleting project template",
+			fmt.Sprintf("Could not delete project template with ID %s: %s", state.ID.ValueString(), err.Error()),
+		)
+		return
+	}
+
+	state.ID = types.StringNull()
```
sheensantoscapadngan, line 434: no need for this, Terraform will clean up the resource entirely

docs/resources/project_template.md
```diff
@@ -0,0 +1,123 @@
...
+  auth = {
+    universal = {
+      client_id     = "<machine-identity-client-id>"
+      client_secret = "<machine-identity-client-secret>"
+    }
+  }
+}
+
+resource "infisical_project_template" "example-project" {
+  name        = "example-project"
```
sheensantoscapadngan, line 36: should be "example-project-template" right?

PR #163, opened 2025-08-12
Return the path of the folder resource (not it's parent folder)

x032205:


DanielHougaard:
@greptile please roast this PR so I don't have to

DanielHougaard (approved):

PR #182, opened 2025-09-01
feat(crossplane): infisical_secret resource support

DanielHougaard:
Added ID field to infisical_secret


DanielHougaard:
Greptile comments are coming from our infisical_secret resource. We should look into resolving most of the above for Terraform

sheensantoscapadngan (approved):

PR #190, opened 2025-09-22
feat(resource/infisical_secret): write-only support

varonix0:
This PR adds write-only support for the `infisical_secret` value. A new parameter called `value_wo` has been added, which serves as a write-only input for the secret value.

When a write-only value is provided the value never gets saved to Terraform state. This is useful for highly sensitive information such as secrets.

Ephemeral secrets are useful for reading secrets for use in a Terraform state, and write-only is useful for creating secrets without saving state.

Note: This only works on Terraform version 1.11.0 and above, hence the go mod updates. Read more here: https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments

### DIFF
--- a/internal/provider/resource/secret_resource.go
+++ b/internal/provider/resource/secret_resource.go
@@ -772,7 +772,6 @@ func (r *secretResource) Update(ctx context.Context, req resource.UpdateRequest,
 
 		// No need to set workspace ID as it is already set in the plan
 		//plan.WorkspaceId = plan.WorkspaceId
-		plan.LastUpdated = types.StringValue(time.Now().Format(time.RFC850))
 
 	} else {
 		resp.Diagnostics.AddError(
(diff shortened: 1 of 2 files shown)


internal/provider/resource/secret_resource.go
```diff
@@ -223,7 +325,7 @@ func (r *secretResource) Create(ctx context.Context, req resource.CreateRequest,
 		}
 
 		// encrypt value
-		encryptedValue, err := crypto.EncryptSymmetric([]byte(plan.Value.ValueString()), plainTextWorkspaceKey)
+		encryptedValue, err := crypto.EncryptSymmetric([]byte(secretData.Value), plainTextWorkspaceKey)
```
maidul98, line 328: q: why did we update this?

maidul98 (approved):
left one q. if it isn't concerning we can merge it

internal/provider/resource/secret_resource.go
DanielHougaard: We use the `getSecretValue()` function to get the secret value everywhere for consistency

PR #207, opened 2025-10-24
feat: gateway as reviewer support

varonix0:
Added support for gateway as reviewer for identity k8s auth. Depends on: https://github.com/Infisical/infisical/pull/4741

I also added a missing example file, and added support for terraform imports.

### DIFF
--- a/internal/client/model.go
+++ b/internal/client/model.go
@@ -1567,8 +1567,8 @@ type CreateIdentityKubernetesAuthResponse struct {
 type UpdateIdentityKubernetesAuthRequest struct {
 	IdentityID              string                         `json:"identityId"`
 	KubernetesHost          *string                        `json:"kubernetesHost"`
-	CACERT                  string                         `json:"caCert,omitempty"`
-	TokenReviewerJwt        string                         `json:"tokenReviewerJwt,omitempty"`
+	CACERT                  string                         `json:"caCert"`
+	TokenReviewerJwt        *string                        `json:"tokenReviewerJwt"`
 	AllowedNamespaces       string                         `json:"allowedNamespaces"`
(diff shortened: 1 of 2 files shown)


internal/client/model.go
```diff
@@ -1562,16 +1566,18 @@ type CreateIdentityKubernetesAuthResponse struct {
 
 type UpdateIdentityKubernetesAuthRequest struct {
 	IdentityID              string                         `json:"identityId"`
-	KubernetesHost          string                         `json:"kubernetesHost"`
-	CACERT                  string                         `json:"caCert"`
-	TokenReviewerJwt        string                         `json:"tokenReviewerJwt"`
+	KubernetesHost          *string                        `json:"kubernetesHost"`
+	CACERT                  string                         `json:"caCert,omitempty"`
+	TokenReviewerJwt        string                         `json:"tokenReviewerJwt,omitempty"`
```
sheensantoscapadngan: by making it omitempty doesn't this mean that you wouldn't be able to unset these since it's for the PATCH request?

internal/provider/resource/identity_kubernetes_auth.go
```diff
@@ -197,6 +229,83 @@ func updateKubernetesAuthStateByApi(ctx context.Context, diagnose diag.Diagnosti
...
+			return
+		}
+
+		if plan.TokenReviewerJWT.ValueString() == "" {
+			diagnose.AddError(
+				"Token reviewer JWT is required",
+				"Token reviewer JWT is required when token reviewer mode is api. Please provide a valid token reviewer JWT.",
+			)
+			return
+		}
```
sheensantoscapadngan: also, isn't this optional because we do self-review without this? 

image

PR #217, opened 2025-11-25
[ENG-4227] fix: adds empty env project template

gupta-piyush19:
adds empty env project template

### DIFF
--- a/internal/client/model.go
+++ b/internal/client/model.go
@@ -2515,11 +2515,11 @@ type Permission struct {
 }
 
 type CreateProjectTemplateRequest struct {
-	Name         string         `json:"name"`
-	Description  string         `json:"description"`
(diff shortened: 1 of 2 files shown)


internal/client/model.go
```diff
@@ -2540,7 +2540,7 @@ type UpdateProjectTemplateRequest struct {
 	Description  string        `json:"description"`
 	Type         string        `json:"type"`
 	Roles        []Role        `json:"roles,omitempty"`
-	Environments []Environment `json:"environments,omitempty"`
+	Environments []Environment `json:"environments"`
```
sheensantoscapadngan: I think with this, we don't have to make it computed anymore?

sheensantoscapadngan (approved):

PR #225, opened 2025-12-11
feature: add PKI resources for profiles, policies, cert authorities and cert authority certificates

carlosmonastyrski:
This PR adds PKI support to the Infisical Terraform provider, enabling organizations to manage their entire certificate lifecycle through Infrastructure as Code. The implementation introduces internal CAs (root and intermediate), external CA integrations (ACME and ADCS), certificate templates and certificate enrollment profiles.

New Resources
### infisical_cert_manager_internal_ca_root
Manages internal root certificate authorities.

### infisical_cert_manager_internal_ca_intermediate
Manages intermediate CAs that chain from root or other intermediate CAs to support hierarchical PKI designs.

###  infisical_cert_manager_ca_certificate
Manages CA certificate generation

### infisical_cert_manager_external_ca_acme
Integrates with ACME providers (e.g., Let’s Encrypt) for automated issuance.

### infisical_cert_manager_external_ca_adcs
Integrates with Microsoft ADCS for enterprise certificate issuance flows.

### infisical_cert_manager_certificate_template
Defines certificate policies: subject rules, SAN constraints, key usages, algorithms, and validity settings.

### infisical_cert_manager_certificate_profile
Enrollment profiles supporting API, EST, and ACME flows with CA-issued or self-signed certificate options.

### DIFF
--- a/docs/resources/cert_manager_internal_ca.md
+++ b/docs/resources/cert_manager_internal_ca.md
@@ -80,6 +80,6 @@ resource "infisical_cert_manager_internal_ca" "issuing" {
 Import is supported using the following syntax:
 
 ```shell
-# This will import the internal CA by project slug and CA ID
-terraform import infisical_cert_manager_internal_ca.example <project_slug>:<ca_id>
+# This will import the internal CA by its ID
+terraform import infisical_cert_manager_internal_ca.example <ca_id>
 ```
--- a/examples/resources/infisical_cert_manager_internal_ca/import.sh
+++ b/examples/resources/infisical_cert_manager_internal_ca/import.sh
@@ -1,2 +1,2 @@
-# This will import the internal CA by project slug and CA ID
-terraform import infisical_cert_manager_internal_ca.example <project_slug>:<ca_id>
+# This will import the internal CA by its ID
+terraform import infisical_cert_manager_internal_ca.example <ca_id>
--- a/internal/provider/resource/cert_manager_internal_ca.go
+++ b/internal/provider/resource/cert_manager_internal_ca.go
@@ -459,15 +459,45 @@ func (r *certManagerInternalCAResource) Delete(ctx context.Context, req resource
 }
 
 func (r *certManagerInternalCAResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
-	parts := strings.Split(req.ID, ":")
-	if len(parts) != 2 {
+	if !r.client.Config.IsMachineIdentityAuth {
 		resp.Diagnostics.AddError(
-			"Invalid import ID",
-			"Import ID must be in the format 'project_slug:ca_id'",
+			"Unable to import CA",
+			"Only Machine Identity authentication is supported for this operation",
 		)
 		return
 	}
 
-	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("project_slug"), parts[0])...)
-	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), parts[1])...)
+	ca, err := r.client.GetInternalCA(infisical.GetCARequest{
+		CAId: req.ID,
+	})
+
+	if err != nil {
+		if err == infisical.ErrNotFound {
+			resp.Diagnostics.AddError(
+				"Error importing CA",
+				fmt.Sprintf("CA with ID %s not found", req.ID),
+			)
+			return
+		}
+		resp.Diagnostics.AddError(
+			"Error importing CA",
+			"Couldn't read CA from Infisical, unexpected error: "+err.Error(),
+		)
+		return
+	}
+
+	project, err := r.client.GetProjectById(infisical.GetProjectByIdRequest{
+		ID: ca.ProjectId,
+	})
+
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error importing CA",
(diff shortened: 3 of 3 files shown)


docs/resources/cert_manager_certificate_template.md
```diff
@@ -0,0 +1,213 @@
...
+
+<a id="nestedblock--key_usages"></a>
+### Nested Schema for `key_usages`
+
+Optional:
+
+- `allowed` (List of String) List of allowed key usages
+- `denied` (List of String) List of denied key usages
+- `required` (List of String) List of required key usages
+
```
sheensantoscapadngan: list out the possible values for this

internal/provider/resource/cert_manager_internal_ca_intermediate.go
```diff
@@ -0,0 +1,496 @@
...
+		Slug: state.ProjectSlug.ValueString(),
+	})
+
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error reading project",
+			"Couldn't read project from Infisical, unexpected error: "+err.Error(),
+		)
+		return
+	}
```
sheensantoscapadngan: might not need this get project call anymore

internal/client/ca_certificate.go
```diff
@@ -0,0 +1,50 @@
...
+		return GenerateCACertificateResponse{}, errors.NewGenericRequestError(operationGenerateCACertificate, err)
+	}
+
+	if response.IsError() {
+		return GenerateCACertificateResponse{}, errors.NewAPIErrorWithResponse(operationGenerateCACertificate, response, nil)
+	}
+
+	return certResponse, nil
+}
+
```
sheensantoscapadngan: IGNORE PLACEMENT OF COMMENT

I defined the parent CA ID in Terraform but it doesn't look like it got linked behind the scenes? 

image
image
image

internal/client/ca_certificate.go
```diff
@@ -0,0 +1,50 @@
...
+	response, err := client.Config.HttpClient.
+		R().
+		SetResult(&certResponse).
+		SetHeader("User-Agent", USER_AGENT).
+		Get(fmt.Sprintf("api/v1/cert-manager/ca/internal/%s/certificate", request.CaId))
+
+	if err != nil {
+		return GetCACertificateResponse{}, errors.NewGenericRequestError(operationGetCACertificate, err)
+	}
+
```
sheensantoscapadngan: IGNORE PLACEMENT OF COMMENT

Wait why do I get this error when trying to create a CA certificate resource for an intermediate CA? 
image

I was trying to do the following via TF:
image

examples/resources/infisical_cert_manager_certificate_profile/resource.tf
```diff
@@ -0,0 +1,193 @@
...
+  enrollment_type = "est"
+  issuer_type     = "ca"
+
+  est_config {
+    passphrase                      = var.est_passphrase
+    disable_bootstrap_ca_validation = false
+    ca_chain                        = var.encrypted_ca_chain
+  }
+}
+
```
sheensantoscapadngan: IGNORE PLACEMENT OF COMMENT

I think we should add some sort of validation so that users don't accidentally configure the key usages and extended key usages field like the ff:

image

The problem is that issuing a certificate from the profile becomes problematic because of the conflict:
image

examples/resources/infisical_cert_manager_certificate_profile/resource.tf
```diff
@@ -0,0 +1,193 @@
...
+  api_config {
+    auto_renew        = true
+    renew_before_days = 14
+  }
+
+  external_configs {
+    template = "WebServerTemplate" # ADCS template name
+  }
+}
+
```
sheensantoscapadngan: IGNORE PLACEMENT OF COMMENT

There seems to be an infinite-drift with the passphrase field?
image

internal/provider/resource/cert_manager_ca_certificate.go
```diff
@@ -0,0 +1,328 @@
...
+
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	certificate, err := r.client.GetCACertificate(infisical.GetCACertificateRequest{
+		CaId: state.CaId.ValueString(),
+	})
+	if err != nil {
```
sheensantoscapadngan: shouldn't we be checking specifically for ErrNotFound here? Looks like at the moment, it will remove the resource as a result of ANY error

internal/provider/resource/cert_manager_ca_certificate.go
```diff
@@ -0,0 +1,328 @@
...
+	var state certManagerCACertificateResourceModel
+
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	certificate, err := r.client.GetCACertificate(infisical.GetCACertificateRequest{
+		CaId: state.CaId.ValueString(),
+	})
```
sheensantoscapadngan: wait.. why do we query it just by the CA ID? A CA can have more than one certificate right?

sheensantoscapadngan: Ah looks like this is indeed causing a bug when I have more than one CA certificate 
image

PR #233, opened 2026-01-16
feature: add PKI resources for certificates

carlosmonastyrski:
This PR introduces a new resource for requesting and managing certificates from Infisical certificate profiles. The resource supports both direct field input (common name, SANs, key algorithm, etc.) and CSR-based certificate requests, with automatic polling for asynchronous certificate issuance from external CAs. Certificates are immutable, so any attribute changes trigger resource replacement, and deletion only removes the certificate from Terraform state since certificates cannot be deleted from Infisical.


docs/resources/cert_manager_certificate.md
```diff
@@ -0,0 +1,117 @@
...
+- `ou` (String) The organizational unit (OU) for the certificate
+- `province` (String) The state/province (ST) for the certificate
+- `signature_algorithm` (String) The signature algorithm for the certificate. Supported: RSA-SHA256, RSA-SHA384, RSA-SHA512, ECDSA-SHA256, ECDSA-SHA384, ECDSA-SHA512
+- `timeout_seconds` (Number) Maximum time to wait for certificate issuance in seconds. Defaults to 3600 (1 hour)
+- `ttl` (String) Time to live for the certificate (e.g., '30d', '90d', '1y')
+
+### Read-Only
+
+- `certificate` (String, Sensitive) The issued certificate in PEM format
+- `certificate_chain` (String, Sensitive) The certificate chain in PEM format
```
sheensantoscapadngan: these should not be sensitive I think

PR #261, opened 2026-04-09
feat(data-source): return secret metadata

adilsitos:
Add a new data source to return secret metadata. Uses `viewSecretValue` in query parameter as false to not return the value. 

How to test:
- create a machine identity 
- modify the terraform example 
- Check that the output returns the version and other metadata.


internal/client/secrets.go
```diff
@@ -288,6 +289,34 @@ func (client Client) GetSingleRawSecretByNameV3(request GetSingleSecretByNameV3R
 	return secretsResponse, nil
 }
 
+func (client Client) GetSingleRawSecretMetadataByNameV3(request GetSingleSecretByNameV3Request) (GetSingleRawSecretByNameSecretResponse, error) {
```
varonix0: I think we should just add a `viewSecretValue` option to the existing `GetSingleRawSecretByNameV3` func to cut down on duplicate code. But I do agree that we should be using `viewSecretValue=false` when fetching metadata specifically, so tokens without the `ReadValue` permission can still fetch metadata

internal/provider/datasource/secret_metadata_data_source.go
```diff
@@ -0,0 +1,143 @@
...
+func (d *SecretMetadataDataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Retrieve metadata for a single Infisical secret without exposing the secret value.",
+
+		Attributes: map[string]schema.Attribute{
+			"name": schema.StringAttribute{
+				Description: "The name of the secret to retrieve metadata for.",
+				Required:    true,
+			},
+			"env_slug": schema.StringAttribute{
```
varonix0: I'd rename this to `environment_slug` to avoid using synonyms

internal/provider/datasource/secret_metadata_data_source.go
```diff
@@ -0,0 +1,143 @@
...
+				Description: "The environment slug where the secret resides.",
+				Required:    true,
+			},
+			"workspace_id": schema.StringAttribute{
```
varonix0: Should be `project_id`

internal/provider/datasource/secret_metadata_data_source.go
```diff
@@ -0,0 +1,143 @@
...
+				Description: "The Infisical project ID.",
+				Required:    true,
+			},
+			"folder_path": schema.StringAttribute{
+				Description: "The path to the folder where the secret is located.",
+				Required:    true,
+			},
+			"type": schema.StringAttribute{
```
varonix0: Should be `secret_type` ideally, I think `type` is too ambiguous, curious what you think here as well

internal/provider/datasource/secret_metadata_data_source.go
```diff
@@ -0,0 +1,143 @@
...
+				Description: "The type of the secret (shared or personal). Defaults to shared.",
+				Optional:    true,
+				Computed:    true,
+			},
+			"workspace": schema.StringAttribute{
```
varonix0: Can be removed, it's the same as `project_id`

internal/provider/datasource/secret_metadata_data_source.go
```diff
@@ -0,0 +1,143 @@
...
+				Description: "The workspace ID of the secret.",
+				Computed:    true,
+			},
+			"secret_version": schema.Int64Attribute{
+				Description: "The version number of the secret.",
+				Computed:    true,
+			},
+			"environment": schema.StringAttribute{
```
varonix0: Can be removed, it's the same as `environment_slug `

internal/provider/datasource/secret_metadata_data_source.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	var data SecretMetadataDataSourceModel
+
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	secretType := "shared"
```
varonix0: I think we should just make this the default value on the schema if the user has not defined anything

internal/provider/datasource/secret_metadata_data_source.go
```diff
@@ -0,0 +1,143 @@
...
+	Workspace     types.String `tfsdk:"workspace"`
+	SecretVersion types.Int64  `tfsdk:"secret_version"`
+	Environment   types.String `tfsdk:"environment"`
+}
+
+func (d *SecretMetadataDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_secret_metadata"
+}
+
+func (d *SecretMetadataDataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
```
varonix0: Right now the data source doesn't actually return metadata. [secrets in Infisical can contain metadata](https://infisical.com/docs/api-reference/endpoints/secrets/read#response-secret-secret-metadata), which makes the resource name misleading. I think we should [also include tags](https://infisical.com/docs/api-reference/endpoints/secrets/read#response-secret-tags), tags are very widely used.

varonix0 (changes requested):
few comments, general feedback would be to analyze the goal of what we're building to see if it makes sense in a broader scope / for more use-cases, and if it makes sense from a user perspective

internal/provider/datasource/secret_metadata_data_source.go
adilsitos: @varonix0 all the new ones should be project_id? I left as workspace_id because the olds ones follow this pattern.

internal/provider/datasource/secret_metadata_data_source.go
adilsitos: This data source schema does not support default.

internal/provider/datasource/secret_metadata_data_source.go
adilsitos: Changed to be `secret_properties_data_source` what do you think? Also included the tags and the metadata.

internal/client/secrets.go
```diff
@@ -259,7 +265,12 @@ func (client Client) UpdateRawSecretV3(request UpdateRawSecretByNameV3Request) e
 	return nil
 }
 
-func (client Client) GetSingleRawSecretByNameV3(request GetSingleSecretByNameV3Request) (GetSingleRawSecretByNameSecretResponse, error) {
+func (client Client) GetSingleRawSecretByNameV3(request GetSingleSecretByNameV3Request, viewSecretValue *bool) (GetSingleRawSecretByNameSecretResponse, error) {
```
adilsitos, line 235: decided to add a parameter here, so we don't have to change the request and change several places. I was also afraid of this staying `hidden` and people calling it and forgetting to define the `viewSecretValue` as true. 

I remember there is a library in go that I can define the defaul on the struct, from what I remember defining the tag would work, but I did a test and this failed (it was always false, even defining `default:"true"`)

internal/client/secrets.go
```diff
@@ -116,33 +116,6 @@ func (client Client) UpdateSecretsV3(request UpdateSecretByNameV3Request) error
 	return nil
 }
 
-func (client Client) GetSingleSecretByNameV3(request GetSingleSecretByNameV3Request) (GetSingleSecretByNameSecretResponse, error) {
```
adilsitos, line 119: this is unused, so I remove it.

varonix0 (approved):

PR #264, opened 2026-04-14
fix(secret-approval): add unorderedListModifier

adilsitos:
Add unorderedListModifier to make sure that the plan does not cause diff in case the elements just change their position. `unorderedListModifier` checks that all elements are equal (from the state and from plan) to make sure a diff is not created without being really needed.  

## How to test 
- go into `examples/resources/infisical_secret_approval_policy` 
- create an approval policy there using `plan` and `apply`. Use more than one environment slug.
- Check the plan again. Check that there is no difference

### DIFF
--- a/internal/provider/resource/secret_approval_policy.go
+++ b/internal/provider/resource/secret_approval_policy.go
@@ -207,14 +207,11 @@ func (r *secretApprovalPolicyResource) Create(ctx context.Context, req resource.
 	}
 
 	var environments []string
+	var environment string
 	if !plan.EnvironmentSlugs.IsNull() {
 		environments = infisicaltf.StringListToGoStringSlice(ctx, resp.Diagnostics, plan.EnvironmentSlugs)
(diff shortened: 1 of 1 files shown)


victorvhs017 (approved):
All good with the PR, only added a commit to fix another issue in the same file that I caught on my tests:

**Bug:** When `environment_slugs` (plural) is used in the Create call, the provider also populates the deprecated `environment` (singular) field with the first element. The backend resolves this with `(environment ? [environment] : environments)`, so when both are present, the singular field wins and only one environment is persisted — the rest are silently dropped.

**Fix:** The Create method now exclusively populates one or the other:
- `environment_slugs` is set → only `environments` is sent, `environment` remains empty
- `environment_slug` (deprecated) is set → only `environment` is sent, `environments` remains nil

This matches what the Update method already does, and ensures the backend correctly receives the full `environments` array when the plural field is used.

PR #266, opened 2026-04-15
fix: retry folder creation after environment creation

thekev:
Creating a folder immediately after creating an environment can fail with a 404 "Failed to retrieve path for folder" even though the environment exists. The folder may actually be created server-side despite the error, causing the next attempt to get a 400 "already exists".

Retry up to 5 times on the 404. On 400 "already exists", recover by looking up the existing folder.

This solves my problem, but I'm not sure it's the best approach. My theory is it's due to read replica delay.

### DIFF
--- a/internal/client/secret_folder.go
+++ b/internal/client/secret_folder.go
@@ -4,7 +4,9 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 	"terraform-provider-infisical/internal/errors"
+	"time"
 )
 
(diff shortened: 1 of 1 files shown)


thekev:
I think what's transpiring is the folder create call checks the provided parent id, but I suspect it's using the read replica. I'm seeing this type of behavior in multiple places. Perhaps the better fix is not here in the provider, but API server side: write ops that validate inputs through a db query should always use the primary writer.

PR #267, opened 2026-04-20
fix: retry 429 with backoff

### LINKED ISSUE #96: Prevent rate limiting from breaking terraform
I have a self hosted instance of infisical and the rate limiting prevent the terraform provider to work.

I understand the limit for the infisical server but it shouldn't prevent Terraform to work, just make the plan and apply take longer.

Is that possible to add a way to retry in case of rate limit ?

sethyates:
## Summary

Fixes #96. Rate-limited (`429`) responses currently bubble straight out as `CallGet... Unsuccessful response ... status-code=429 ... "Rate limit exceeded. Please try again in 58 seconds"`, which aborts the Terraform run mid-plan/apply. This PR makes the shared HTTP client retry on rate-limit and transient server errors, honoring the server's `Retry-After` hint when given.

Example error this used to produce:

```
Couldn't read project secret folder from Infisical, unexpected error:
CallGetSecretFolderByID Unsuccessful response
[GET https://app.infisical.com/api/v1/folders/<id>]
[status-code=429] [request-id=...]
[message="Rate limit exceeded. Please try again in 58 seconds"]
```

## Changes

- New `internal/client/retry.go` with `configureRetries(*resty.Client)`:
  - Retries on `429 Too Many Requests`.
  - Does **not** retry on other 4xx (auth/validation failures still fail fast).
  - Wait time sourced in this order:
    1. `Retry-After` header — supports both RFC 7231 forms (delta-seconds like `58`, and HTTP-date like `Wed, 21 Oct 2015 07:28:00 GMT`).
    2. `X-RateLimit-Reset` — supports delta-seconds and absolute Unix timestamps.
    3. Fallback: jittered exponential backoff, 1s → 60s (resty's built-in `jitterBackoff`).
  - Up to 5 retries per request. At 60s cap per retry this bounds worst-case added latency to ~5 minutes, well under typical Terraform operation timeouts.
- `NewClient` calls `configureRetries` on the resty client before any auth or API call, so **every** endpoint (auth/login, secrets, folders, identities, projects, integrations, etc.) picks up the behavior through the single shared `Config.HttpClient`.
- Tests in `internal/client/retry_test.go`:
  - Unit tests for `parseRetryAfter` (integer seconds, HTTP-date future/past, negative values, garbage input) and `parseRateLimitReset` (delta seconds, absolute Unix timestamp, negatives, garbage).
  - Integration test with an `httptest.Server` that returns `429` with `Retry-After: 1` twice then `200` — verifies exactly 3 handler calls, >= 2s elapsed (proof the header is honored, not ignored).
  - Integration test asserting `400` responses are **not** retried (regression guard).

## Why this approach

- One configure site, not per-call. All SDK operations use the same resty client, so installing retry policy on `Config.HttpClient` fixes every current and future call site without touching the individual resource/data-source files.
- Uses resty's native retry primitives (`SetRetryCount`, `SetRetryWaitTime`, `SetRetryMaxWaitTime`, `AddRetryCondition`, `SetRetryAfter`) rather than hand-rolling a loop, keeping the surface small and behaviorally consistent with the rest of the ecosystem.
- `SetRetryAfter` returning `0` delegates to resty's jittered exponential backoff (see `sleepDuration` in resty/v2), which avoids thundering-herd when many resources retry concurrently.

## Non-goals / things deliberately not changed

- No change to public provider schema or configuration — retry tuning is not user-exposed. Happy to add `provider "infisical" { ... max_retries = N }` knobs in a follow-up if maintainers want them; defaults seemed safer to ship first.
- No change to error surfaces. If all retries exhaust, the same `NewAPIErrorWithResponse` error the user saw before is returned — so existing error handling and logging keep working.
- Network-level errors (DNS, connection reset) are **not** retried here. Terraform already retries entire operations at the graph level, and retrying network errors at the HTTP layer risks masking genuine misconfiguration. Happy to broaden if maintainers prefer.

## Test plan

- [x] `go build -v .`
- [x] `go generate ./...` — no diff
- [x] `go test ./internal/client/...` — all retry tests pass (2.5s)
- [ ] `make testacc` — requires live Infisical creds; not run locally, CI will cover
- [ ] Manual repro: plan against a project large enough to hit rate limiting (the original issue scenario). Before this PR the run failed immediately with the 429 above; after, it completes.

Closes https://github.com/Infisical/terraform-provider-infisical/issues/96

### DIFF
--- a/internal/client/retry.go
+++ b/internal/client/retry.go
@@ -1,0 +1,104 @@
+package infisicalclient
+
+import (
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/go-resty/resty/v2"
+)
+
+const (
+	defaultRetryCount       = 5
+	defaultRetryWaitTime    = 1 * time.Second
+	defaultRetryMaxWaitTime = 60 * time.Second
+)
+
+// configureRetries installs retry behavior on the resty client so transient
+// errors (notably 429 Too Many Requests) do not fail the Terraform run.
+// The Retry-After header is honored when present; otherwise jittered
+// exponential backoff is used, capped at defaultRetryMaxWaitTime.
+func configureRetries(c *resty.Client) {
+	c.SetRetryCount(defaultRetryCount)
+	c.SetRetryWaitTime(defaultRetryWaitTime)
+	c.SetRetryMaxWaitTime(defaultRetryMaxWaitTime)
+
+	c.AddRetryCondition(func(r *resty.Response, err error) bool {
+		if r == nil {
+			return false
+		}
+		switch r.StatusCode() {
+		case http.StatusTooManyRequests,
+			http.StatusBadGateway,
+			http.StatusServiceUnavailable,
+			http.StatusGatewayTimeout:
+			return true
+		}
+		return false
+	})
+
+	c.SetRetryAfter(func(_ *resty.Client, r *resty.Response) (time.Duration, error) {
+		if r == nil {
+			return 0, nil
+		}
+		if d, ok := parseRetryAfter(r.Header().Get("Retry-After")); ok {
+			return d, nil
+		}
+		if d, ok := parseRateLimitReset(r.Header().Get("X-RateLimit-Reset")); ok {
+			return d, nil
+		}
+		return 0, nil
+	})
+}
+
+// parseRetryAfter parses an HTTP Retry-After header. It supports both the
+// delta-seconds and HTTP-date forms defined by RFC 7231.
+func parseRetryAfter(v string) (time.Duration, bool) {
+	v = strings.TrimSpace(v)
+	if v == "" {
+		return 0, false
+	}
+	if secs, err := strconv.Atoi(v); err == nil {
+		if secs < 0 {
+			secs = 0
+		}
+		return time.Duration(secs) * time.Second, true
+	}
+	if t, err := http.ParseTime(v); err == nil {
+		d := time.Until(t)
+		if d < 0 {
+			d = 0
+		}
+		return d, true
+	}
+	return 0, false
+}
+
+// parseRateLimitReset parses the X-RateLimit-Reset header, which some APIs
+// send instead of (or alongside) Retry-After. The value is either a
+// delta-seconds integer or an absolute Unix timestamp; both are supported.
+func parseRateLimitReset(v string) (time.Duration, bool) {
+	v = strings.TrimSpace(v)
+	if v == "" {
+		return 0, false
+	}
+	n, err := strconv.ParseInt(v, 10, 64)
+	if err != nil {
+		return 0, false
+	}
+	if n < 0 {
+		return 0, true
+	}
+	// Values above ~year 2001 in unix seconds are absolute timestamps;
+	// smaller numbers are delta-seconds.
+	const absoluteTimestampThreshold = 1_000_000_000
+	if n >= absoluteTimestampThreshold {
+		d := time.Until(time.Unix(n, 0))
+		if d < 0 {
+			d = 0
+		}
+		return d, true
+	}
+	return time.Duration(n) * time.Second, true
+}
--- a/internal/client/retry.go
+++ b/internal/client/retry.go
@@ -29,10 +29,7 @@ func configureRetries(c *resty.Client) {
 			return false
 		}
 		switch r.StatusCode() {
-		case http.StatusTooManyRequests,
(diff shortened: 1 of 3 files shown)


fmarrero:
@sethyates thanks for your work on this.. would love to see it merged!

internal/client/retry.go
```diff
@@ -0,0 +1,104 @@
...
+
+	c.AddRetryCondition(func(r *resty.Response, err error) bool {
+		if r == nil {
+			return false
+		}
+		switch r.StatusCode() {
+		case http.StatusTooManyRequests,
+			http.StatusBadGateway,
+			http.StatusServiceUnavailable,
+			http.StatusGatewayTimeout:
```
adilsitos: Let's retry only on 429 (`statusTooManyRequests`). It is possible that some endpoint is not idempotent, and a 5xx error might process something and create duplicates in case we retry. The `429` will be the safest option for us.

internal/client/retry_test.go
adilsitos, line 1: only the provider test is executed during the CI. We could extend that to also run this test, but I believe this one is not necessary now. We can remove it.

internal/client/retry.go
```diff
@@ -0,0 +1,104 @@
...
+	// Values above ~year 2001 in unix seconds are absolute timestamps;
+	// smaller numbers are delta-seconds.
+	const absoluteTimestampThreshold = 1_000_000_000
+	if n >= absoluteTimestampThreshold {
+		d := time.Until(time.Unix(n, 0))
+		if d < 0 {
+			d = 0
+		}
+		return d, true
+	}
```
adilsitos, line 90: This can be:
```suggestion
		d := max(time.Until(time.Unix(n, 0)), 0)
		return d, true
```

internal/client/retry.go
```diff
@@ -0,0 +1,104 @@
...
+			secs = 0
+		}
+		return time.Duration(secs) * time.Second, true
+	}
+	if t, err := http.ParseTime(v); err == nil {
+		d := time.Until(t)
+		if d < 0 {
+			d = 0
+		}
+		return d, true
```
adilsitos: this can be: 


```suggestion
		d = max(d, 0)
		return d, true
```

adilsitos (changes requested):
Thanks for your contribution! I just left some comments to make this more resilient.

adilsitos:
@greptile review this

internal/client/retry.go
```diff
@@ -0,0 +1,104 @@
...
+		if d, ok := parseRetryAfter(r.Header().Get("Retry-After")); ok {
+			return d, nil
+		}
+		if d, ok := parseRateLimitReset(r.Header().Get("X-RateLimit-Reset")); ok {
+			return d, nil
+		}
+		return 0, nil
+	})
+}
+
```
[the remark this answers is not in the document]
adilsitos: Not required, the default approach that we have on the code right now is intended.

internal/client/retry_test.go
```diff
@@ -0,0 +1,126 @@
...
+	rc.SetBaseURL(server.URL)
+	configureRetries(rc)
+
+	_, err := rc.R().Get("/")
+	if err != nil {
+		t.Fatalf("request error: %v", err)
+	}
+	if got := atomic.LoadInt32(&count); got != 1 {
+		t.Fatalf("handler calls = %d, want 1 (no retries on 400)", got)
+	}
```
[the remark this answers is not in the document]
adilsitos: Not required.

internal/client/retry.go
sethyates: Done

internal/client/retry_test.go
sethyates: Are you suggesting to remove the test?

internal/client/retry.go
sethyates: Done

internal/client/retry.go
sethyates: Done

sethyates:
@adilsitos addressed the feedback

internal/client/retry_test.go
adilsitos: Sorry. Actually, this is fine. We can leave the test. I will add a change on the CI so we can have this as a requirement on the CI.

sethyates: OK.

adilsitos (approved):

sethyates:
Thanks!

PR #269, opened 2026-04-23
fix(secret-approval-policy): Define environments as empty slice 

adilsitos:
When the deprecated value is used, we need to define the environments as empty instead of nil, so we don't break the api contract.

If the user sends a `slugs` and later send a `slug`, this can break the update. Changed the provider to also support this.


internal/provider/resource/secret_approval_policy.go
```diff
@@ -212,6 +212,7 @@ func (r *secretApprovalPolicyResource) Create(ctx context.Context, req resource.
 		environments = infisicaltf.StringListToGoStringSlice(ctx, resp.Diagnostics, plan.EnvironmentSlugs)
 	} else {
 		environment = plan.EnvironmentSlug.ValueString()
+		environments = []string{}
```
adilsitos, line 215: I didn't want to define this at the creation to avoid other errors with append that could happen. I this case this would not happen, but who know in the future.

victorvhs017 (dismissed):

victorvhs017 (approved):

PR #272, opened 2026-04-28
feat(app-connection): add support for vault

adilsitos:
Add a new app connection so it is possible to create a vault app connection using terraform.


varonix0:
@claude give this a review

varonix0 (approved):

PR #276, opened 2026-05-05
improvement(resource-folder): add description

adilsitos:
add description into folder resource. 

How to test: 
- Create a new machine identity 
- Modify the example that creates folder 
- Add description and check if the folder was created in infisical and it has the defined description. 
- Also test update an delete (using terraform)


mathnogueira (approved):
Tested locally and it works fine. 

Tested: create/update/delete/outputs

PR #278, opened 2026-05-14
chore: add retry to aws app connection op

Thiago-AS:


internal/provider/resource/app_connection/app_connection_aws.go
```diff
@@ -152,6 +153,28 @@ func NewAppConnectionAwsResource() resource.Resource {
...
+		// visible to STS immediately, causing AssumeRole calls to fail.
+		IsRetryableError: func(err error) bool {
+			if err == nil {
+				return false
+			}
+
+			retryableMessages := []string{
+				"sts:assumerole",
+				"cannot be assumed",
+				"is not authorized to perform: sts:assumerole",
```
victorvhs017: "sts:assumerole" is a substring of "is not authorized to perform: sts:assumerole", so the last one will never be validated

victorvhs017 (changes requested):

victorvhs017 (approved):

PR #279, opened 2026-05-14
fix: prevent perpetual diff on secrets with multiple tags

thekev:
When `tag_ids` in state has a different order than what config produces, the framework detects a value change and marks all Computed attributes with null config values as unknown. The UnorderedList modifier on `tag_ids` runs after this and suppresses the tag diff, but `last_updated` (Computed, not in config) is already marked "(known after apply)". This triggers an update plan, which also surfaces value_wo as a side effect. 

This changes `tag_ids` from List to Set so ordering is ignored. This solves the root cause. The last_updated attribute is also updated to reflect the API's updatedAt field, but that is just extra credit now with the Set change.

### DIFF
--- a/internal/provider/resource/secret_resource.go
+++ b/internal/provider/resource/secret_resource.go
@@ -6,9 +6,7 @@ import (
 	"fmt"
 	"strings"
 	infisical "terraform-provider-infisical/internal/client"
-	modifiers "terraform-provider-infisical/internal/pkg/modifiers"
 	pkg "terraform-provider-infisical/internal/pkg/strings"
-	"time"
 
 	"github.com/hashicorp/go-uuid"
 	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
@@ -55,7 +53,7 @@ type secretResourceModel struct {
 	ValueWOVersion types.Int64     `tfsdk:"value_wo_version"`
 	WorkspaceId    types.String    `tfsdk:"workspace_id"`
 	LastUpdated    types.String    `tfsdk:"last_updated"`
-	Tags           types.List      `tfsdk:"tag_ids"`
+	Tags           types.Set       `tfsdk:"tag_ids"`
 	Metadata       types.Map       `tfsdk:"metadata"`
 	ID             types.String    `tfsdk:"id"`
 }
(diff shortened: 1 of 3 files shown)


thekev:
Repro case:

1. Create an `infisical_secret` resource with 2+ `tag_ids`.
2. Swap the tag order in terraform.tfstate (e.g. ["alpha_id", "beta_id"] → ["beta_id", "alpha_id"])
3. `tofu plan -refresh=false` → shows perpetual diff on `last_updated` and `value_wo`

Different ordering in state could come from before UnorderedList was added in 0.16.12. Dunno exactly how that managed to get reordered, but this fixes our problem.

internal/provider/resource/secret_resource.go
```diff
@@ -588,7 +593,6 @@ func (r *secretResource) Update(ctx context.Context, req resource.UpdateRequest,
 		plan.Metadata = types.MapNull(types.StringType)
 	}
 
-	plan.LastUpdated = types.StringValue(time.Now().Format(time.RFC850))
```
mathnogueira, line 591: On line 572, if you make UpdateRawSecretV3 return the updated secret(s), you can actually use the `UpdatedAt` property here. 

Otherwise, the `LastUpdated` property is never updated after valid secret updates.

Reproduction:

* apply new secret and get `LastUpdated` property
* apply any change
* Check `LastUpdated` timestamp and notice it's same same as when we created first applied the secret.

on line 572:   
```go
updatedSecret, err := r.client.UpdateRawSecretV3(updateRequest)
```

later:
```go
if updatedSecret.UpdatedAt != "" {
	plan.LastUpdated = types.StringValue(updatedSecret.UpdatedAt)
}
```

mathnogueira (commented):
Hey Kevin, first of all. Thanks for contributing to Infisical.

Only a small edge case I found while testing your solution. Left a description on how to reproduce and possible fix.

mathnogueira (changes requested):

internal/provider/resource/secret_resource.go
thekev: I believe this is addressed in the latest push. The Set change addresses the root cause, and the timestamp is now just extra credit. I have not seen unwanted deltas due to the type change. 

I tested the timestamp change by editing a tag on a secret outside of terraform, then an apply saw the diff, reverted my change, and wrote the new last_updated to state. If I edit the tags but put them back the way they were, an apply describes no changes, but does update the last_updated attribute in state.

mathnogueira:
@thekev I reviewed this and I'm still seeing a weird behavior. When you apply the changes and right after reapply, the `UpdatedAt` field diverges and causes terraform to see it as a change. 

It only happens once though.

thekev:
> @thekev I reviewed this and I'm still seeing a weird behavior. When you apply the changes and right after reapply, the `UpdatedAt` field diverges and causes terraform to see it as a change.
> 
> It only happens once though.

Does it only happen when the provider changes from released to this PR? I'm not sure I see a clear way around it. The `time.Now()` change to API field will likely see a diff. 

Would you rather I:
1. Revert to `time.Now()` and drop that change. The `Set` change still fixes the root cause of our perpetual diff.
2. Remove the `last_updated` attribute entirely. It doesn't reflect reality without the API time. 
3. Find a way to implement the change to API time without the extra diff you're seeing?

mathnogueira:
Well, this is only happening once. I think this PR already improves the behavior we have. I can create a follow up PR next week to solve this issue.

Thanks for contributing to it @thekev

mathnogueira (approved):

PR #288, opened 2026-06-02
improvement(app-connections): add support for gateways

adilsitos:
This changes the base app connection file to allow the creation of app connections. 

Test: 
- start a vault instance locally 
- try to create an app connection using the resource.tf example from the vault app connection 
- First create it without gateway 
- Update it (add the gateway)
- remove the gateway 
- Make sure that the changes are being reflected inside infisical.


adilsitos:
@greptile can you check again and update the summary

adilsitos:
@greptile can you check again and update the summary

PR #298, opened 2026-06-16
feat(resource): add adopt_existing flag to infisical_project_identity to adopt auto-created admin

stefan-fast:
## Problem
When a machine identity creates a project, Infisical automatically adds that identity as an admin member. Any subsequent `infisical_project_identity` resource targeting the same identity will fail with "Identity is already a member". This makes it impossible to manage the role assignment of the auto-added identity through Terraform without a manual terraform import.

## Changes
`internal/provider/resource/project_identity_resource.go`:  
- Added an optional `adopt_existing` boolean attribute (default: `false`) to `infisical_project_identity`
- When `adopt_existing = true`, Create checks for an existing membership before attempting to create:
  - Membership found: Adopts it and updates roles to the desired state
  - Not a member: Creates normally like before
  - Unexpected error: Fails with a diagnostic error
- When `adopt_existing = false` (the default), behaviour is unchanged and it tries to add the identity to project, failing when it already is a member
- Refactored duplicated role-building and API-to-state mapping logic out of `Create`, `Read`, `Update`, and `ImportState` into shared helpers `buildRolesFromPlan` and `mapAPIRolesToState`
- Added missing `ResourceWithImportState` compile-time interface assertion (the `ImportState` method was already implemented)

`docs/resources/project_identity.md`:   
- Documents the new `adopt_existing` attribute

## Note
The default is `false` to preserve existing behaviour and stay consistent with the [HashiCorp recommendation](https://developer.hashicorp.com/terraform/plugin/framework/resources/create) that `Create` should return an error when a resource already exists. The opt-in flag is intended for the specific case where the membership is created implicitly by the platform and cannot be prevented.


internal/provider/resource/project_identity_resource.go
```diff
@@ -226,62 +306,72 @@ func (r *ProjectIdentityResource) Create(ctx context.Context, req resource.Creat
...
+			Roles:      createRoles,
+		})
+		if err != nil {
+			resp.Diagnostics.AddError(
+				"Error attaching identity to project",
+				"Couldn't create project identity to Infiscial, unexpected error: "+err.Error(),
+			)
+			return
+		}
 	}
```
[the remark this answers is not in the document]
stefan-fast: The TOCTOU window here is theoretical rather than practical, I think. The scenario where `adopt_existing = true` is used is: the provisioning identity creates the project, Infisical auto-adds it as admin, and then `infisical_project_identity` needs to update the role. By the time `Create` runs, the identity is already a member. The pre-flight `GET` detects this and takes the `PATCH` path correctly.

The remaining race (`GET` returns not-found, then something adds the identity before `POST`) would, as mentioned here, require a concurrent external process modifying the same project identity membership in the same short window. In this use case that's unexpected and, if it happens, an error would probably be an appropriate outcome. Something is modifying the project in parallel and the operator should be informed.

Also I want to note here:
Catching the "Identity is already a member" error to silently fall through to Update would require string-matching against an 400 response message (`BadRequestError` maps to HTTP 400 with no unique error code for this case). That couples the provider to a backend string that could change without notice, creating a silent failure mode worse than the theoretical race it prevents.

internal/provider/resource/project_identity_resource.go
```diff
@@ -226,62 +306,72 @@ func (r *ProjectIdentityResource) Create(ctx context.Context, req resource.Creat
...
+		default:
+			resp.Diagnostics.AddError(
+				"Error checking project identity membership",
+				"Couldn't check existing membership, unexpected error: "+getErr.Error(),
+			)
+			return
+		}
+	}
+
+	if alreadyMember {
```
victorvhs017, line 366: Since adoption takes over a membership this config didn't create and rewrites its roles, let's emit a warning so there's a trace in the apply output.


```suggestion
	if alreadyMember {
		 resp.Diagnostics.AddWarning(
           "Adopted existing project identity membership",
           fmt.Sprintf("Identity %s was already a member of project %s (membership %s); the existing membership was adopted into Terraform state and its roles were updated to match the configuration (previous roles: %s).",
                plan.IdentityID.ValueString(), plan.ProjectID.ValueString(), existing.Membership.ID, <previous role slugs>),
    )
```

docs/resources/project_identity.md
```diff
@@ -57,6 +57,10 @@ resource "infisical_project_identity" "test-identity" {
 - `project_id` (String) The id of the project
 - `roles` (Attributes List) The roles assigned to the project identity (see [below for nested schema](#nestedatt--roles))
 
+### Optional
+
+- `adopt_existing` (Boolean) When true, if the identity is already a member of the project (e.g. auto-added by Infisical when the project was created by this identity), the existing membership is adopted and its roles are updated to match the desired state instead of returning an error. Defaults to false.
+
```
victorvhs017, line 79: Docs here are generated by tfplugindocs (go generate ./...), so please put the behavior text in the schema Description and regenerate rather than editing the .md directly.

Please also add an `adopt_existing = true` example (with a comment about the creator-membership use case) to `examples/resources/infisical_project_identity/resource.tf `so it lands in the generated docs.

victorvhs017 (changes requested):
Thanks for this @stefan-fast!

One blocker we have here beside both my comments is the merge conflicts, please also check those.

stefan-fast:
> Thanks for this @stefan-fast!
> 
> One blocker we have here beside both my comments is the merge conflicts, please also check those.

Appreciated the feedback, @victorvhs017! I’ve applied the changes and rebased onto the latest `main`. Ready for another look.

internal/provider/resource/project_identity_resource.go
```diff
@@ -329,17 +341,75 @@ func (r *ProjectIdentityResource) Create(ctx context.Context, req resource.Creat
...
+			}
+			prevSlugs = append(prevSlugs, slug)
+		}
+		previousRoles := strings.Join(prevSlugs, ", ")
+		if previousRoles == "" {
+			previousRoles = "(none)"
+		}
+		resp.Diagnostics.AddWarning(
+			"Adopted existing project identity membership",
+			fmt.Sprintf("Identity %s was already a member of project %s (membership %s); the existing membership was adopted into Terraform state and its roles were updated to match the configuration (previous roles: %s).",
```
victorvhs017:
```suggestion
			fmt.Sprintf("Identity %s was already a member of project %s (membership %s); the existing membership will be adopted into Terraform state and its roles updated to match the configuration (previous roles: %s).",
```

victorvhs017 (changes requested):
Just a nit on my wording on the message, because this is before the update. But everything else looks good!

victorvhs017 (approved):

victorvhs017:
@stefan-fast this is deployed in the [version 0.19.4 of the Infisical Terraform Provider ](https://registry.terraform.io/providers/Infisical/infisical/latest/docs/resources/project_identity)

Thank you for your contribution!

stefan-fast:
Amazing, thank you too!

PR #305, opened 2026-06-22
feat(project_user): add import 

adilsitos:
- create a new project and add new users 
- try to import the state using `terraform import` (check the docs added on this PR) 
- validate that it is possible to add it 
- check use cases like: id is not UUID, path traversal, UUID does not exist


adilsitos:
@greptile check again and update the summary

mathnogueira (approved):

PR #312, opened 2026-06-30
feat: add project-scoped identity resource and related examples

victorvhs017:
## Add `infisical_project_scoped_identity` (with role support)

Replaces the hidden `infisical_project_identity_provisioning` resource with a properly named, documented resource that also manages project roles at create time.

### Breaking change

`infisical_project_identity_provisioning` is removed. Migrate to `infisical_project_scoped_identity` and add a required `roles` block (≥1 permanent role). Import: `<project_id>,<identity_id>`.

### Highlights

- Atomic create: identity + roles in one API call
- Role updates via project membership endpoint
- Shared role helpers + validation hardened for `infisical_project_identity` too
- Docs and example added; client renamed to `ProjectScopedIdentity*`

### Requires

Infisical backend that accepts `roles` on `POST /projects/{id}/identities`.

### Test plan

- [ ] Create (built-in, custom, temporary, multi-role) — no drift
- [ ] Update roles in place
- [ ] Import single/multi-role identity
- [ ] Validation errors (duplicate slug, all-temporary, bad timestamp)


internal/provider/resource/project_scoped_identity_resource.go
```diff
@@ -0,0 +1,546 @@
...
+
+// Metadata returns the resource type name.
+func (r *ProjectScopedIdentityResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_project_scoped_identity"
+}
+
+// Schema defines the schema for the resource.
+func (r *ProjectScopedIdentityResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Create and manage a machine identity scoped to a single project in Infisical, including its project role(s). Project-scoped identities are bound to one project and cannot be assigned to other projects. Only Machine Identity authentication is supported for this resource.",
```
adilsitos, line 55: on the other resource `project_identity_resource` we have the following description: 

```
		Description: "Create project identities & save to Infisical. Only Machine Identity authentication is supported for this data source",
```

maybe we can change that description, so users know that they cannot create a machine identity inside a project there. Maybe we can even point them to this resource.

internal/provider/resource/project_scoped_identity_resource.go
```diff
@@ -0,0 +1,546 @@
...
+	elements := make([]attr.Value, len(methods))
+	for i, m := range methods {
+		elements[i] = types.StringValue(m)
+	}
+	return types.ListValueMust(types.StringType, elements)
+}
+
+// metadataFromAPI converts API metadata into the Terraform model representation, returning an
+// empty (non-nil) slice when there is no metadata to keep plan/state consistent.
+func metadataFromAPI(apiMetadata []infisical.MetaEntry) []MetaEntry {
```
adilsitos, line 537: nit: not sure if it is required, but isn't it possible that this can stay in a different way because of position? I'm thinking in a scenario where the returned value from the API is different only with key position. Maybe we could sort this?

adilsitos (commented):
added two minor comments, let me know if they make sense.

internal/provider/resource/project_scoped_identity_resource.go
victorvhs017: No drift in this case because we are using SetNestedAttribute for the metadata, this expects the results to be unordered.

adilsitos (approved):

PR #317, opened 2026-07-07
feat: add basic_constraints and extra subject types

carlosmonastyrski:
Extends the certificate policy resource so it covers the newer policy options the backend already supports: control over whether issued certificates can act as certificate authorities, more subject fields (org unit, state, locality), and clearer docs on how values are matched.

### DIFF
--- a/internal/provider/resource/cert_manager_certificate_policy.go
+++ b/internal/provider/resource/cert_manager_certificate_policy.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	infisical "terraform-provider-infisical/internal/client"
 
+	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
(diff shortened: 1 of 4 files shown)


internal/provider/resource/cert_manager_certificate_policy.go
```diff
@@ -124,7 +152,7 @@ func (r *certManagerCertificatePolicyResource) Schema(_ context.Context, _ resou
 		},
 		Blocks: map[string]schema.Block{
 			"subject": schema.ListNestedBlock{
-				Description: "Subject attribute policies for the certificate policy",
+				Description: "Subject attribute policies for the certificate policy. Each block constrains a single subject DN attribute (e.g. common_name, organization). Values are matched against the corresponding attribute parsed from the CSR; the '*' wildcard matches any sequence of characters (including dots). For common_name, matching uses the CN attribute only, so domainComponent (DC) attributes are ignored.",
```
sheensantoscapadngan, line 155: > For common_name, matching uses the CN attribute only, so domainComponent (DC) attributes are ignored

wait what do you mean by this last sentence? why do we explicitly call this out?

sheensantoscapadngan (approved):

PR #324, opened 2026-07-14
feat: add domain_component to PKI resources

carlosmonastyrski:


sheensantoscapadngan (approved):

sheensantoscapadngan (changes requested):
empty array value for domain_components is not working... this might be desired if I want to overwrite the default as an example...
Image

sheensantoscapadngan (approved):

PR #326, opened 2026-07-16
feat: access_approval_policy and secret_approval_policy

mathnogueira:
Adds:

* secret_approval_policy for crossplane

Updated:

* access_approval_policy in crossplane and terraform to add the missing fields
* secret_approval_policy for terraform, adding the missing fields

### DIFF
--- a/crossplane/access_approval_policy_resource/access_approval_policy_resource.go
+++ b/crossplane/access_approval_policy_resource/access_approval_policy_resource.go
@@ -17,14 +17,17 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
+// NewAccessApprovalPolicyResource is a helper function to simplify the provider implementation.
 func NewAccessApprovalPolicyResource() resource.Resource {
 	return &accessApprovalPolicyResource{}
 }
 
+// accessApprovalPolicyResource is the resource implementation.
 type accessApprovalPolicyResource struct {
 	client *infisical.Client
 }
 
+// accessApprovalPolicyResourceModel describes the data source data model.
 type accessApprovalPolicyResourceModel struct {
 	ID                    types.String `tfsdk:"id"`
 	ProjectID             types.String `tfsdk:"project_id"`
@@ -42,10 +45,12 @@ type accessApprovalPolicyResourceModel struct {
 	RequestExpirationTime types.String `tfsdk:"request_expiration_time"`
 }
 
+// Metadata returns the resource type name.
 func (r *accessApprovalPolicyResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
 	resp.TypeName = req.ProviderTypeName + "_access_approval_policy"
 }
 
+// Schema defines the schema for the resource.
 func (r *accessApprovalPolicyResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
 	resp.Schema = schema.Schema{
 		Description: "Create access approval policy for your projects",
@@ -127,6 +132,7 @@ func (r *accessApprovalPolicyResource) Schema(_ context.Context, _ resource.Sche
 	}
 }
 
+// Configure adds the provider configured client to the resource.
 func (r *accessApprovalPolicyResource) Configure(_ context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
 	if req.ProviderData == nil {
 		return
@@ -146,6 +152,7 @@ func (r *accessApprovalPolicyResource) Configure(_ context.Context, req resource
 	r.client = client
 }
 
+// Create creates the resource and sets the initial Terraform state.
 func (r *accessApprovalPolicyResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
 	if !r.client.Config.IsMachineIdentityAuth {
 		resp.Diagnostics.AddError(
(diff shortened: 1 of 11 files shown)


crossplane/access_approval_policy_resource/access_approval_policy_resource.go
```diff
@@ -288,45 +340,82 @@ func (r *accessApprovalPolicyResource) Read(ctx context.Context, req resource.Re
...
-	state.RequiredApprovals = types.Int64Value(accessApprovalPolicy.AccessApprovalPolicy.RequiredApprovals)
-	state.EnforcementLevel = types.StringValue(accessApprovalPolicy.AccessApprovalPolicy.EnforcementLevel)
+	policy := accessApprovalPolicy.AccessApprovalPolicy
 
-	groupApprovers := make([]string, 0)
-	userApprovers := make([]string, 0)
+	if policy.DeletedAt != nil {
+		resp.State.RemoveResource(ctx)
+		return
+	}
```
mathnogueira, line 282: Weird enough, our API is returning deleted rows. Didn't open a PR to solve this on the API because that could be a breaking change.

mathnogueira:
@greptile review it once more

mathnogueira:
`approvals_required` block cannot be removed in the server side without changing the approvers, so the concern from greptile is not applicable.

crossplane/access_approval_policy_resource/resource.tf
```diff
@@ -24,15 +24,51 @@ resource "infisical_access_approval_policy" "prod-policy" {
...
+      step     = 2
+    },
+    {
+      type = "group"
+      id   = "83d5cf3b-3580-4aaf-872e-564a8ccaaf86"
+      step = 2
+    },
   ]
 
+  # Option 2 (DEPRECATED): Use "group_approvers" and "user_approvers" flat lists.
```
varonix0: I don't think we want to introduce something that has been deprecated to Crossplane, we should probably get rid of this before we introduce it to Crossplane

crossplane/secret_approval_policy_resource/secret_approval_policy_resource.go
```diff
@@ -0,0 +1,517 @@
...
+		return
+	}
+
+	secretApprovalPolicy, err := r.client.GetSecretApprovalPolicyByID(infisical.GetSecretApprovalPolicyByIDRequest{
+		ID: state.ID.ValueString(),
+	})
+
+	if err != nil {
+		if err == infisical.ErrNotFound {
+			resp.State.RemoveResource(ctx)
```
varonix0, line 242: Just a comment for knowledge share purposes; its super important to remove the resource from state if the request returns 404 not found, because if we throw an error in read, crossplane will keep retrying the read function instead of creating the resource

docs/resources/access_approval_policy.md
```diff
@@ -68,10 +91,15 @@ resource "infisical_access_approval_policy" "prod-policy" {
...
+- `approvals_required` (Attributes List) The number of approvals required per step for multi-step approval policies (see [below for nested schema](#nestedatt--approvals_required))
+- `bypassers` (Attributes Set) The bypassers who can bypass the approval policy (see [below for nested schema](#nestedatt--bypassers))
 - `enforcement_level` (String) The enforcement level of the policy. This can either be hard or soft
 - `environment_slug` (String) (DEPRECATED, Use environment_slugs instead) The environment to apply the access approval policy to
 - `environment_slugs` (List of String) The environments to apply the access approval policy to
+- `max_time_period` (String) The maximum time period for the access approval, specified as a duration string (e.g. '1h', '30m', '2d'). Use 'permanent' or leave empty for no limit.
 - `name` (String) The name of the access approval policy
+- `request_expiration_time` (String) The time after which the access request expires, specified as a duration string (e.g. '1h', '3d', '72h'). Must be between 1 minute and 1 year. Use 'never' or leave empty for no expiration.
```
varonix0: I think we should document what the default is since this is optional

crossplane/access_approval_policy_resource/access_approval_policy_resource.go
```diff
@@ -12,61 +11,41 @@ import (
...
+	GroupBypassers        types.List   `tfsdk:"group_bypassers"`
+	UserBypassers         types.List   `tfsdk:"user_bypassers"`
+	RequiredApprovals     types.Int64  `tfsdk:"required_approvals"`
+	EnforcementLevel      types.String `tfsdk:"enforcement_level"`
+	AllowSelfApproval     types.Bool   `tfsdk:"allow_self_approval"`
+	MaxTimePeriod         types.String `tfsdk:"max_time_period"`
+	RequestExpirationTime types.String `tfsdk:"request_expiration_time"`
 }
 
-// Metadata returns the resource type name.
```
varonix0, line 53: Should not be removed, common pattern to have these for all resources, same for the schema below

crossplane/access_approval_policy_resource/access_approval_policy_resource.go
```diff
@@ -190,7 +127,6 @@ func (r *accessApprovalPolicyResource) Schema(_ context.Context, _ resource.Sche
 	}
 }
 
-// Configure adds the provider configured client to the resource.
```
varonix0, line 114: Seems like we removed these everywhere

varonix0 (dismissed):

docs/resources/access_approval_policy.md
mathnogueira: > Use 'permanent' or leave empty for no limit.
> Use 'never' or leave empty for no expiration.

Do you think we should make them something like this?

> By default, access is permanent
> By default, it never expires

akhilmhdh: If omitted, the default value is `false` - this is how aws adds it

akhilmhdh (approved):

PR #330, opened 2026-07-22
fix:  crossplane build

mathnogueira:
## Problem

`internal/provider/resource/project_scoped_identity_resource.go` was using shared functions from `internal/provider/resource/project_identity_resource.go`, however, that file is replaced by `scripts/prepare-crossplane.sh`, removing those functions and replacing a type that was being used.

This would cause go to fail to build due to those functions being missing. This duplicates the functions and the type renaming them to prevent conflicts.

### DIFF
--- a/internal/pkg/terraform/identity_roles.go
+++ b/internal/pkg/terraform/identity_roles.go
@@ -1,0 +1,150 @@
+package terraform
+
+import (
+	"fmt"
+	"sort"
+	infisical "terraform-provider-infisical/internal/client"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+const TemporaryModeRelative = "relative"
+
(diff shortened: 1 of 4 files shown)


victorvhs017 (changes requested):
Instead of duplicating the role helpers here to avoid Crossplane break, can we extract them into `internal/pkg/terraform `the same way we did for access approval?

`prepare-crossplane.sh` overwrites `project_identity_resource.go`, so sharing helpers from that file breaks the Crossplane build. A shared package under i`nternal/pkg/` is never overwritten, so both Terraform resources can call it safely.

This is the access approval file: `internal/pkg/terraform/access_approval.go` it creates the `ValidateAndMapApprovers` function that is used inside `internal/provider/resource/access_approval_policy_resource.go`

victorvhs017 (approved):

PR #332, opened 2026-07-24
feat: add resource for AWS Certificate Manager sync

carlosmonastyrski:


carlosmonastyrski:
@greptile please re-review one more time and update the Greptile Summary above

internal/provider/resource/pki_sync/pki_sync_aws_certificate_manager.go
```diff
@@ -0,0 +1,133 @@
...
+	"can_remove_certificates": types.BoolType,
+	"include_root_ca":         types.BoolType,
+	"preserve_arn":            types.BoolType,
+}
+
+func NewPkiSyncAwsCertificateManagerResource() resource.Resource {
+	return &PkiSyncBaseResource{
+		App:              infisical.PkiSyncAppAWSCertificateManager,
+		SyncName:         "AWS Certificate Manager",
+		ResourceTypeName: "_pki_sync_aws_certificate_manager",
```
saifsmailbox98: do you think we should use the name "certificate sync" and not "pki sync" in everything user facing?

internal/provider/resource/pki_sync/base_pki_sync.go
```diff
@@ -0,0 +1,377 @@
...
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, plan)...)
+}
+
+// hydrateStateFromApi refreshes model in place from the API's current view of the sync, keyed
+// by model.ID. Create and Update call it so state reflects any server-side normalization
+// (e.g. trimmed name/schema) instead of the raw plan. It returns false when the sync no longer
+// exists.
+func (r *PkiSyncBaseResource) hydrateStateFromApi(ctx context.Context, model *PkiSyncBaseResourceModel) (bool, diag.Diagnostics) {
```
saifsmailbox98: is there any field that this sets that's not already in the plan? If it is only for normalization then I don't think it works, because TF rejects a normalized value for a required attribute.

internal/provider/resource/pki_sync/base_pki_sync.go
```diff
@@ -0,0 +1,377 @@
...
+		}
+		diags.AddError(
+			"Error reading PKI sync",
+			"Couldn't read PKI sync, unexpected error: "+err.Error(),
+		)
+		return false, diags
+	}
+
+	model.ConnectionID = types.StringValue(pkiSync.ConnectionID)
+	model.Name = types.StringValue(pkiSync.Name)
```
saifsmailbox98: name is required and the backend can trim it. So if a config value has trailing whitespace it will keep showing a diff on every refresh.

Can we use customtypes.TrimmedStringType here like we do in `identity_kubernetes_auth.go`?

internal/provider/resource/certificate_sync/certificate_sync_aws_certificate_manager.go
```diff
@@ -0,0 +1,133 @@
...
+		ResourceTypeName: "_pki_sync_aws_certificate_manager",
+		AppConnection:    infisical.AppConnectionAppAWS,
+		DestinationConfigAttributes: map[string]schema.Attribute{
+			"aws_region": schema.StringAttribute{
+				Required:    true,
+				Description: "The AWS region to sync certificates to (e.g. us-east-1).",
+			},
+		},
+		SyncOptionsAttributes: map[string]schema.Attribute{
+			"certificate_name_schema": schema.StringAttribute{
```
saifsmailbox98, line 52: This also is required and can get trimmed on the backend. Can we use `customtypes.TrimmedStringType` here as well?

The attr type map entry will also need to change from `types.StringType`

saifsmailbox98 (changes requested):

saifsmailbox98 (approved):

PR #334, opened 2026-07-27
feat(provider-auth): define auth by only using env variables

adilsitos:
Before this change, when the auth brackets were not defined on the schema the auth method was using a fallback to use a token auth method. With this change the users can define the method they want to use, and define on the env vars the required credentials. 

It was added a new schema attribute called `auth_method `, where the user can define which method wants to use (either by defining it on the TF file or by using the env var `INFISICAL_AUTH_METHOD`).
### Scenario 1: Auth brackets and auth method defined on the TF file
- expected: fail during plan. it is only possible to define one way of authentication

### Scenario 2: Auth brackets and auth method  defined on the env vars 
- auth brackets take precedence, so they are the ones that are used 

### Scenario 3: Auth method defined on the TF file and credentials on env vars 
- Works without any error. 

### Scenario 4: Everything defined on the env vars 
- Authentication works using only the env vars


adilsitos:
@greptile review again and update the summary

internal/provider/provider.go
```diff
@@ -192,6 +205,13 @@ func (p *infisicalProvider) Schema(ctx context.Context, _ provider.SchemaRequest
 					},
 				},
 			},
+			"auth_method": schema.StringAttribute{
```
varonix0: Hmm, why do we set this on the provider? We already have the `auth` object which holds all the auth credentials. Having the auth method env var + the env vars for identity ID, client secret, etc etc makes sense, but can you help clarify why we need it on the provider schema itself? Feels like its conflicting with the `auth` object

varonix0 (changes requested):

internal/provider/provider.go
adilsitos: my thought here was that if we want to use the env vars we need to define it in a way, so I thought this could be on the env var or on the schema itself, giving an option for the user.  I imagine that when you have this on the schema it is more clear that the the env var is being used (maybe this name is not the best to be honest). 

But I can remove it and just leave on the env var if it makes more sense.

varonix0: I'm thinking we can remove it from the schema for now. We can easily add it in, in the future without breaking anything. However if we wanted to remove it from he schema in the future, that will be tough without introducing breaking changes

varonix0 (changes requested):
I think we should write the provider doc more like this. It wasn't clear to me when reading it how to actually set it up and use the auth methods via env vars:


```go
# Authenticate via environment variables instead of the auth block:
#
# provider "infisical" {}
#
#
# Authenticate via Universal Auth:
# export INFISICAL_AUTH_METHOD="universal"
# export INFISICAL_UNIVERSAL_AUTH_CLIENT_ID="<machine-identity-client-id>"
# export INFISICAL_UNIVERSAL_AUTH_CLIENT_SECRET="<machine-identity-client-secret>"
#
# Authenticate via Token Auth:
# ...
#
# Authenticate via OIDC Auth:
# ...
#
# Authenticate via Kubernetes Auth:
# ...
#
# Authenticate via AWS IAM Auth:
# ...

```

varonix0 (approved):

PR #342, opened 2026-08-04
feat: add infisical_alert resource

Thiago-AS:
Adds a new `infisical_alert` resource for Infisical's Alerts API. An alert watches a single resource for a single event and sends notifications to one or more channels.

Right now it supports `resource_type = "identity.authentication"` with an expiry condition, so you can notify users *N* days before a machine identity's auth credentials expire. The condition block is named after the condition itself, while `resource_type` defines what is expiring. Adding support for more resource types or events is just a matter of extending `alertEvents`.

Channels are defined as a map keyed by any name you choose. Each channel contains exactly one config block, which determines its type (`email`, `slack`, `webhook`, or `pagerduty`). The key is only used locally and is never sent to Infisical. Renaming the channel updates it in place and preserves its delivery history. Changing the key replaces the channel.

## Testing locally

If you already have a `dev_overrides` entry for `infisical/infisical` in `~/.terraformrc`, skip `terraform init` since it'll fail. Just go straight to `validate`, `plan`, and `apply`.

### 2. Build the provider

```sh
go build -o terraform-provider-infisical .
```

### 3. Create some test resources

In the local UI:

- Create a machine identity with **Identity: Edit + Read** permissions.
- Grab its Universal Auth client ID and client secret.
- Create another machine identity to use as the `resource_id`.
- Grab a user ID for the email channel by calling:

### 4. Apply the example

```sh
export INFISICAL_HOST=http://localhost:8080
export INFISICAL_UNIVERSAL_AUTH_CLIENT_ID=...
export INFISICAL_UNIVERSAL_AUTH_CLIENT_SECRET=...
```

A good starting point is `examples/resources/infisical_alert/resource.tf`

```text
terraform {
  required_providers {
    infisical = {
      source = "infisical/infisical"
    }
  }
}

provider "infisical" {
  auth = {
    universal = {}
  }
}

variable "org_id" {
  type        = string
  description = "Organization the test objects are created in. Decode the JWT from POST /api/v1/auth/universal-auth/login to find it."
  default     = "e2a8b185-e6c8-4eea-b45a-4226085d35ee"
}

variable "recipient_user_id" {
  type        = string
  description = "User to notify by email. Set to null to skip the email channel entirely. GET /api/v2/organizations/<org>/memberships, use the inner user.id, not the membership id."
  default     = "242b53db-952c-44c1-a44a-1eb942948fcf"
}

resource "infisical_identity" "watched" {
  name   = "alert-e2e-watched"
  org_id = var.org_id
  role   = "no-access"
}

resource "infisical_identity_universal_auth" "watched" {
  identity_id = infisical_identity.watched.id
}

resource "infisical_identity_universal_auth_client_secret" "watched" {
  identity_id = infisical_identity_universal_auth.watched.identity_id
  description = "expires inside the alert window"
  ttl         = 10 * 24 * 60 * 60
}

resource "infisical_alert" "org_scoped" {
  resource_type = "identity.authentication"
  resource_id   = infisical_identity.watched.id
  name          = "alert-e2e org scoped"
  description   = "Rotate the watched identity's client secret before it expires"

  expiry = {
    alert_before_days = 30
  }

  channels = merge(var.recipient_user_id == null ? {} : {
    security_team = {
      name = "Security team"
      email = {
        recipients = [{ type = "user", id = var.recipient_user_id }]
      }
    }
  })

  depends_on = [infisical_identity_universal_auth_client_secret.watched]
}

output "org_alert" {
  value = {
    id          = infisical_alert.org_scoped.id
    org_id      = infisical_alert.org_scoped.org_id
    watching    = infisical_alert.org_scoped.resource_id
    channel_ids = { for key, channel in infisical_alert.org_scoped.channels : key => channel.id }
  }
}
```

Then walk through the lifecycle:

- Run `terraform plan` immediately after `apply`. It should report **no changes**. This catches secret round-trip drift.
- Rename the alert and a channel, toggle `enabled`, and change `alert_before_days`. Everything should update in place and channel IDs should stay the same.
- Remove `description` and verify it becomes `null`. Run `terraform plan` again to make sure there's no drift.
- Change a channel's type while keeping the same key. The channel ID should be replaced without hitting an inconsistent result error.
- Run:
  ```sh
  terraform state rm <addr>
  terraform import <addr> <alert-id>
  ```
  Channels should be imported using their Infisical names as keys, and the only diff should be the write-only secrets.
- Delete the alert manually:
  Then run `terraform plan`. It should detect the missing resource and plan to recreate it.
- Finally, remove the alert, run:
  ```sh
  terraform destroy
  ```
  Then confirms that the alert has been deleted.


internal/provider/resource/alert/alert_channels.go
```diff
@@ -0,0 +1,689 @@
...
+				"webhook": schema.SingleNestedAttribute{
+					Optional:    true,
+					Description: "Sends notifications to an HTTPS endpoint as a CloudEvents payload.",
+					Attributes: map[string]schema.Attribute{
+						"url": schema.StringAttribute{
+							Required:    true,
+							Description: "The HTTPS URL the CloudEvents payload is sent to.",
+							Validators: []validator.String{
+								infisicaltf.HttpsUrlValidator,
+							},
```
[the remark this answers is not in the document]
Thiago-AS: The Alerts API already restricts webhook destinations independently, and it does so at the point that actually matters

internal/provider/resource/alert/alert_channels.go
```diff
@@ -0,0 +1,689 @@
...
+				"id": schema.StringAttribute{
+					Computed:    true,
+					Description: "The ID of the channel in Infisical.",
+					PlanModifiers: []planmodifier.String{
+						stringplanmodifier.UseStateForUnknown(),
+					},
+				},
+				"name": schema.StringAttribute{
+					Required:    true,
+					Description: "The name the channel is shown under in Infisical. Can be changed freely, and has to be unique within the alert.",
```
victorvhs017: The backend allows duplicate names for channels. Restricting it here would cause issues with imported alerts that have duplicate names in the channels.

docs/resources/alert.md
```diff
@@ -0,0 +1,227 @@
...
+      client_id     = "<machine-identity-client-id>"
+      client_secret = "<machine-identity-client-secret>"
+    }
+  }
+}
+
+# Email the security team 30 days before an organization level machine identity's
+# authentication credentials expire
+resource "infisical_alert" "ci_identity" {
+  resource_type = "identity.authentication"
```
victorvhs017, line 38: We need a list in the documentation with all available resource types; we only have one now, but users don't know that.

An existing example is permissions: https://infisical.com/docs/internals/permissions/project-permissions#subject-role

docs/resources/alert.md
```diff
@@ -0,0 +1,227 @@
...
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `channels` (Attributes Map) The channels the alert is delivered to, keyed by a name of your choosing. The key is what identifies a channel to Terraform and is never sent to Infisical, so renaming a channel updates it in place and it keeps the deliveries it has already made. Changing a key, on the other hand, deletes the channel and creates a new one, so the new channel notifies about everything that is still expiring, even if the old one already did. Each channel carries exactly one configuration block, and that block is what gives the channel its type. At least one and at most 10 channels are allowed. (see [below for nested schema](#nestedatt--channels))
+- `name` (String) The name of the alert.
+- `resource_id` (String) The ID of the resource to watch. For the identity.authentication resource type this is a machine identity's ID.
+- `resource_type` (String) The type of the resource to watch. Options: identity.authentication.
```
victorvhs017, line 121: Then we can link the docs here.

internal/provider/resource/alert/alert_channels.go
```diff
@@ -0,0 +1,689 @@
...
+}
+
+func sortedMapKeys(elements map[string]attr.Value) []string {
+	keys := make([]string, 0, len(elements))
+	for key := range elements {
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
+	return keys
+}
```
victorvhs017: nit: Both functions are the same; we could use a generic type, or even better, use `slices.Sorted(maps.Keys(channels))` that returns the same result.

internal/client/alert.go
```diff
@@ -0,0 +1,112 @@
...
+
+const (
+	operationCreateAlert  = "CallCreateAlert"
+	operationGetAlertByID = "CallGetAlertByID"
+	operationUpdateAlert  = "CallUpdateAlert"
+	operationDeleteAlert  = "CallDeleteAlert"
+)
+
+func isDuplicateAlertResponse(response *resty.Response) bool {
+	return response.StatusCode() == http.StatusBadRequest && strings.Contains(response.String(), "already exists")
```
victorvhs017: Nit: Not a fan of comparing texts, especially in another repository. If, for some reason, the text changes or someone adds this same text in another error unrelated to the one expected here, it would break the check.

internal/provider/resource/alert/alert.go
victorvhs017, line 1: If we add a second channel to the alert, the apply fails. Here's the example I used; apply this and then uncomment step 2 and apply again.

```
terraform {
  required_providers {
    infisical = { source = "infisical/infisical" }
  }
}

provider "infisical" {
  host = "http://localhost:8080"
  auth = {
    universal = {
      client_id     = "<machine-identity-client-id>"
      client_secret = "<machine-identity-client-secret>"
    }
  }
}

resource "infisical_alert" "repro" {
  resource_type = "identity.authentication"
  resource_id   = "<any-machine-identity-id>"
  name          = "add-channel repro"

  expiry = { alert_before_days = 30 }

  channels = {
    first = {
      name    = "First"
      webhook = { url = "https://example.com/first" }
    }

    # STEP 2: add this block and apply again -> "Provider produced inconsistent
    # result after apply: .channels["second"].id: was null, but now ..."
    # second = {
    #   name    = "Second"
    #   webhook = { url = "https://example.com/second" }
    # }
  }
}

```

victorvhs017 (changes requested):

internal/client/alert.go
Thiago-AS: Yeah, the thing is, we should have unique error codes for that instead of relying on the error message, but this is pretty much the pattern we use in the repo since we don't have that. Will remove and use the list endpoint to compare existence. One more request, but it should not be a problem for now.

docs/resources/alert.md
Thiago-AS: We'll do that once we have more options. Since this isn't a top-level feature for now, we'll keep it as is.

victorvhs017 (approved):
LGTM

PR #344, opened 2026-08-06
support approval request and pool until promise is merged

mathnogueira:
if project has a change policy in place that do no allow machine identities to bypass it, terraform has a bad time syncing the state. 

As a solution we are using the same polling strategy as `cert_manager_certificate`: if an approval is returned from the CRUD endpoint, we start polling the state of the request and only return if:

1. timeout (1h)
2. all secrets are either approved or rejected

This ensures terraform is capable of storing the actual state of the secret in a sync way.


internal/provider/resource/secret_resource.go
```diff
@@ -639,6 +756,106 @@ func (r *secretResource) Delete(ctx context.Context, req resource.DeleteRequest,
...
+}
+
+// pollSecretApprovalRequest polls the secret approval request endpoint until the
+// request is merged, closed, or the context is cancelled. Returns the approval
+// details once merged, or adds a diagnostic error otherwise.
+func (r *secretResource) pollSecretApprovalRequest(ctx context.Context, approvalID string, timeoutSeconds int64, diagnostics *diag.Diagnostics) *infisical.SecretApprovalRequestDetails {
+	timeout := time.Duration(timeoutSeconds) * time.Second
+	startTime := time.Now()
+
+	minInterval := 2 * time.Second
```
adilsitos, line 776: nit: I'm afraid that 2 seconds might be too small to begin with no? Maybe a larger time to avoid start starting too early might be good

internal/provider/resource/secret_resource.go
```diff
@@ -639,6 +756,106 @@ func (r *secretResource) Delete(ctx context.Context, req resource.DeleteRequest,
...
+	if next > maxInterval {
+		return maxInterval
+	}
+	return next
+}
+
+// pollSecretApprovalRequest polls the secret approval request endpoint until the
+// request is merged, closed, or the context is cancelled. Returns the approval
+// details once merged, or adds a diagnostic error otherwise.
+func (r *secretResource) pollSecretApprovalRequest(ctx context.Context, approvalID string, timeoutSeconds int64, diagnostics *diag.Diagnostics) *infisical.SecretApprovalRequestDetails {
```
adilsitos, line 772: maybe before starting this, we could show a warning on the diagnostics? Because I feel like the message we get from terraform CLI is that it is taking long to respond to the request, this might confuse the user, and make then stop the plan. We won't have any problems, but it might be good to avoid noise on support channels and it is also better UX IMO

mathnogueira: Not possible, terraform does not allow you to print anything while running, it only shows messages after the process is complete: either finished or timedout 😢

adilsitos: I was thinking in something like this: 

```
resp.Diagnostics.AddWarning(
		"Secret creation can take up to 30 seconds",
		"New secrets are propagated across replicas before the write is acknowledged. "+
			"The apply will appear to hang; this is expected.",
	)
```

adilsitos: wouldn't it work?

mathnogueira: I'm checking this

mathnogueira: Does not work

CleanShot 2026-08-24 at 14 36 23

CleanShot 2026-08-24 at 14 36 40


It only gets displayed AFTER terraform finishes running:

CleanShot 2026-08-24 at 14 37 15

varonix0 (approved):

PR #347, opened 2026-08-11
fix: explicit sequence and approval steps

mathnogueira:
Fix issue that caused access policies to be created as multistep instead of single step.

### DIFF
--- a/internal/provider/resource/access_approval_policy_resource.go
+++ b/internal/provider/resource/access_approval_policy_resource.go
@@ -221,7 +221,12 @@ func (r *accessApprovalPolicyResource) Create(ctx context.Context, req resource.
 	var approvers []infisical.CreateAccessApprovalPolicyApprover
 	for _, a := range validatedApprovers {
 		approvers = append(approvers, infisical.CreateAccessApprovalPolicyApprover{
-			ID: a.ID, Name: a.Name, Type: a.Type,
+			ID:   a.ID,
+			Name: a.Name,
+			Type: a.Type,
+			// For now we only support approvals with one step
+			// So all approvers are for step 1
+			Sequence: 1,
 		})
 	}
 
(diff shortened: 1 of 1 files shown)


internal/provider/resource/access_approval_policy_resource.go
```diff
@@ -272,13 +277,19 @@ func (r *accessApprovalPolicyResource) Create(ctx context.Context, req resource.
...
+		SecretPath:        plan.SecretPath.ValueString(),
+		Approvers:         approvers,
+		Bypassers:         bypassers,
+		RequiredApprovals: plan.RequiredApprovals.ValueInt64(),
+		ApprovalsRequired: []infisical.AccessApprovalPolicyApprovalsRequired{
+			{
+				NumberOfApprovals: plan.RequiredApprovals.ValueInt64(),
+				StepNumber:        1,
+			},
+		},
```
mathnogueira, line 292: We were missing this one.

internal/provider/resource/access_approval_policy_resource.go
```diff
@@ -272,13 +277,19 @@ func (r *accessApprovalPolicyResource) Create(ctx context.Context, req resource.
...
-		Approvers:             approvers,
-		Bypassers:             bypassers,
-		RequiredApprovals:     plan.RequiredApprovals.ValueInt64(),
+		Name:              plan.Name.ValueString(),
+		ProjectSlug:       projectDetail.Slug,
+		Environments:      environments,
+		SecretPath:        plan.SecretPath.ValueString(),
+		Approvers:         approvers,
+		Bypassers:         bypassers,
+		RequiredApprovals: plan.RequiredApprovals.ValueInt64(),
```
mathnogueira, line 286: Even though it's marked as `optional` in the API, this is actually required. Otherwise we get a 422 error. We can tackle that in a separate PR, I want to unblock the customer.

mathnogueira:
@greptile review again

varonix0 (approved):

PR #351, opened 2026-08-26
feat: datadog api key rotation resource

mathnogueira:
## Context

Add `infisical_secret_rotation_datadog_api_key` as a new resource to manage secret rotations for Datadog API Keys.


internal/provider/resource/secret_rotation/secret_rotation_datadog_api_key.go
```diff
@@ -0,0 +1,130 @@
...
+		ParametersAttributes: map[string]schema.Attribute{
+			"name": schema.StringAttribute{
+				Required:    true,
+				Description: "The name for the generated Datadog API keys.",
+			},
+		},
+		SecretsMappingAttributes: map[string]schema.Attribute{
+			"api_key_id": schema.StringAttribute{
+				Required:    true,
+				Description: "The name of the secret that the Datadog API key ID will be mapped to.",
```
adilsitos, line 39: nit: this shouldn't be `The id of the API key id? `

internal/provider/resource/secret_rotation/secret_rotation_datadog_api_key.go
```diff
@@ -0,0 +1,130 @@
...
+			},
+			"api_key": schema.StringAttribute{
+				Required:    true,
+				Description: "The name of the secret that the rotated Datadog API key will be mapped to.",
+			},
+		},
+
+		ReadParametersFromPlan: func(ctx context.Context, plan SecretRotationBaseResourceModel) (map[string]interface{}, diag.Diagnostics) {
+			parametersMap := make(map[string]interface{})
```
adilsitos, line 48: nit: maybe we could use `any`?

adilsitos (approved):
just two nits

internal/provider/resource/secret_rotation/secret_rotation_datadog_api_key.go
mathnogueira: Oh no, this is correct. This is the mapping setting, so it's indeed the name of the secret that will hold the rotated Datadog API Key ID.

internal/provider/resource/secret_rotation/secret_rotation_datadog_api_key.go
mathnogueira: The `ReadParametersFromPlan` function expects a `map[string]interface{}`. Even though they are equivalent, I'll use the same type just to keep the same signature and prevent confusion.

This pattern is used in other resources.

PR #353, opened 2026-09-01
feat(pki): add digicert app connection and certificate authority resources

saifsmailbox98:
## Context

Adds `infisical_app_connection_digicert` and `infisical_cert_manager_external_ca_digicert`. DigiCert was supported in the API but had no Terraform resource, so CertCentral CAs had to be created by hand.

This targets the **CertCentral Services API** CA (`CaType.DIGICERT`), not DigiCert over ACME. DigiCert-via-ACME is already covered by the existing `infisical_cert_manager_external_ca_acme` resource.

```hcl
resource "infisical_app_connection_digicert" "digicert" {
  name   = "digicert"
  method = "api-key"
  credentials = {
    api_key = var.digicert_api_key
    region  = "us"
  }
}

resource "infisical_cert_manager_external_ca_digicert" "ssl" {
  name              = "digicert-ssl"
  app_connection_id = infisical_app_connection_digicert.digicert.id
  organization_id   = 2146472
  product_name_id   = "ssl_basic"
}
```

The app connection is built on the existing `AppConnectionBaseResource`, so it follows the same shape as the other 23 connections. The CA resource follows `cert_manager_external_ca_adcs`.

### Notes

- `organization_id` is an int and the shared `CertificateAuthorityConfiguration` struct is `omitempty` throughout, so it's stored as `*int`. A plain `int` silently drops the field on the wire.
- `purpose` is Optional+Computed because the backend normalizes an absent value to `ssl`. Same for `status`, matching the other CA resources. Note that leaving `status` out of config means Terraform adopts the server value rather than correcting out-of-band changes.
- `verified_contact` is only required for code signing when the CertCentral org isn't already CS-validated.
- Valid values for `organization_id` and `product_name_id` are account-specific. They come from `GET /app-connections/digicert/:id/organizations` and `/products`.


internal/provider/resource/cert_manager_external_ca_digicert.go
```diff
@@ -0,0 +1,417 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+const digiCertCaPurposeSsl = "ssl"
+
+var (
+	_ resource.Resource = &certManagerExternalCADigiCertResource{}
+
+	SUPPORTED_DIGICERT_CA_PURPOSES = []string{digiCertCaPurposeSsl, "code_signing"}
```
carlosmonastyrski: small nit, one is a plain string and the other one a variable

saifsmailbox98: Valid, made `code_signing` a constant as well so both are consistent.

carlosmonastyrski (approved):

PR #354, opened 2026-09-04
feat: auth0 client secret rotation resource

lobocv:
## Description

Adds `infisical_secret_rotation_auth0_client_secret`, exposing the Auth0 client secret rotation that the Infisical API already supports.

The resource follows the existing rotation pattern: it configures `SecretRotationBaseResource` with the Auth0-specific `parameters` and `secrets_mapping` blocks and the four callbacks that translate between Terraform's snake_case and the API's camelCase. `secret_rotation_azure_client_secret.go` is the closest sibling; Auth0 differs only in taking `client_id` alone as a parameter.

Two client constants are added. `auth0-client-secret` matches `SecretRotation.Auth0ClientSecret` in the API, and `auth0` matches `AppConnection.Auth0`. The app connection constant is new because, unlike Datadog, no Auth0 connection resource exists in the provider yet.

### Usage

```hcl
resource "infisical_secret_rotation_auth0_client_secret" "example" {
  name          = "auth0-client-secret-rotation"
  project_id    = "<project-id>"
  environment   = "<environment-slug>"
  secret_path   = "/"
  connection_id = "<app-connection-id>"

  parameters = {
    client_id = "<auth0-application-client-id>"
  }

  secrets_mapping = {
    client_id     = "AUTH0_CLIENT_ID"
    client_secret = "AUTH0_CLIENT_SECRET"
  }
}
```

The Auth0 app connection itself still has to be created in the UI. Adding an `infisical_app_connection_auth0` resource was deliberately left out of scope.

## Scope

Six files, additions only:

| File | Purpose |
| --- | --- |
| `internal/provider/resource/secret_rotation/secret_rotation_auth0_client_secret.go` | The resource |
| `internal/provider/provider.go` | Registration |
| `internal/client/secret_rotation.go` | `auth0-client-secret` rotation type |
| `internal/client/app_connection.go` | `auth0` app connection type |
| `examples/resources/infisical_secret_rotation_auth0_client_secret/resource.tf` | Example, inlined into the generated doc |
| `docs/resources/secret_rotation_auth0_client_secret.md` | Generated by `go generate ./...` |

## Testing

Tested end to end against Infisical Cloud with a real Auth0 tenant, using a machine-to-machine application authorized for the Management API and a separate throwaway application as the rotation target.

- Create rotates the target application's secret and writes both values into the project.
- `terraform plan` after apply reports no changes, so the plan-to-API and API-to-state callbacks round-trip correctly.
- Update applies cleanly for `rotation_interval` and `rotate_at_utc`.
- Out-of-band changes made in the UI are detected on refresh.
- Destroy removes the rotation.

`go build`, `go vet`, and `gofmt` are clean. `go generate ./...` produces no further diff.

## Pre-existing issues found while testing

None of these are introduced here or specific to Auth0. They affect all eight rotation resources and are listed for follow-up rather than fixed in this PR.

1. **`secret_path`, `environment`, and `connection_id` silently no-op on update.** The API's update schema does not accept them (`secret-rotation-v2-schemas.ts`), and the service reads the location from the existing row. The provider sends them anyway and then writes the plan into state, so a change to any of them produces an apply that reports success, changes nothing, and shows the same diff again on the next plan. These three should carry `RequiresReplace`, as `project_id` already does, and come out of `UpdateSecretRotationRequest`.

2. **No way to revoke credentials or delete secrets on destroy.** The API exposes `revokeGeneratedCredentials` and `deleteSecrets` as query parameters, and the UI offers both as checkboxes. `DeleteSecretRotation` sends neither, so Terraform users cannot opt in. The defaults match the UI defaults, so current behaviour is correct, just not configurable.

3. **Stray debug print.** `base_secret_rotation.go` prints the description on every read via `fmt.Printf`.


internal/provider/resource/secret_rotation/base_secret_rotation.go
```diff
@@ -100,12 +100,14 @@ func (r *SecretRotationBaseResource) Schema(_ context.Context, _ resource.Schema
...
+				Required:      true,
+				Description:   "The slug of the project environment to rotate secrets from.",
+				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
 			},
 			"secret_path": schema.StringAttribute{
-				Required:    true,
-				Description: "The folder path to rotate secrets from.",
+				Required:      true,
+				Description:   "The folder path to rotate secrets from.",
+				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
```
adilsitos, line 110: Since this is forcing the recreation (delete + create), maybe we could add this on the description, and say something like: "The folder path to rotate secrets from. Editing this recreates the rotation"

adilsitos (approved):
LGTM. I just added a docs recommendation. But it is optional

internal/provider/resource/secret_rotation/base_secret_rotation.go
lobocv: I think we should be okay because `terraform plan` and `apply` make it very clear that the resource will be deleted and recreated.

PR #355, opened 2026-09-16
feat: support sign_ra_with_ca in Certificate Manager SCEP profile

andrew951:
## Summary
- Adds `sign_ra_with_ca` to the `scep_config` block on `infisical_cert_manager_application_profile`, so the RA certificate can be signed by the profile's CA instead of self-signed (required for strict SCEP clients like Apple and Microsoft Intune, per the [SCEP enrollment docs](https://infisical.com/docs/documentation/platform/pki/applications/enrollment-methods/scep)).
- Wires the field through the request/response models (`SetPkiApplicationScepEnrollmentRequest.SignRaWithCa`, `PkiApplicationScepEnrollmentState.SignRaWithCa`) and the resource's create/update/refresh logic.
- Regenerated docs and example to include the new attribute.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/...`
- [x] `gofmt -l` clean on changed files
- [ ] Acceptance test against a live Infisical instance that supports `signRaWithCa` on the SCEP enrollment endpoint (not run here — no test environment available)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


andrew951:
Addressed: added a `ModifyPlan` check on this resource that errors during `terraform plan` if `scep_config.sign_ra_with_ca` would change on an already-configured SCEP enrollment, instead of letting Terraform plan an in-place update that the API would reject. The error message tells the user to remove `scep_config` (disable SCEP) in one apply and add it back with the new value in a subsequent apply, matching the documented "disable and reconfigure" requirement.

PR #357, opened 2026-09-17
fix: support sign_ra_with_ca in Certificate Manager SCEP profile

carlosmonastyrski:
- Adds `sign_ra_with_ca` to the `scep_config` block on `infisical_cert_manager_application_profile`, so the RA certificate can be signed by the profile's CA instead of self-signed (required for strict SCEP clients like Apple and Microsoft Intune, per the [SCEP enrollment docs](https://infisical.com/docs/documentation/platform/pki/applications/enrollment-methods/scep)).
- Wires the field through the request/response models (`SetPkiApplicationScepEnrollmentRequest.SignRaWithCa`, `PkiApplicationScepEnrollmentState.SignRaWithCa`) and the resource's create/update/refresh logic.
- Regenerated docs and example to include the new attribute.

Created from https://github.com/Infisical/terraform-provider-infisical/pull/355 by @andrew951, with follow-up fixes.


andrew951 (approved):

internal/provider/resource/cert_manager_application_profile.go
```diff
@@ -248,6 +259,51 @@ func (r *certManagerApplicationProfileResource) Configure(_ context.Context, req
...
+
+	var priorSign, planSign types.Bool
+	resp.Diagnostics.Append(req.State.GetAttribute(ctx, signRaWithCaPath(), &priorSign)...)
+	resp.Diagnostics.Append(req.Plan.GetAttribute(ctx, signRaWithCaPath(), &planSign)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if signRaWithCaChanged(priorSign, planSign) {
+		resp.Diagnostics.AddAttributeWarning(
```
saifsmailbox98: Should this be AddAttributeError? A warning leaves terraform plan at exit 0, which might not be ideal as apply is going to fail, and we already know about it.

saifsmailbox98 (changes requested):

saifsmailbox98 (approved):

